PR #2, opened 2015-01-16
Update README.md with the correct year

jbfp:
Changes the year in the README title from 2014 to 2015 which seems to be the correct year, according to the video title.


stephentoub:
LGTM

davkean:
Sorry for the late reply, LGTM.

PR #4, opened 2015-06-09
Add 2015-06-09-os-version

terrajobst:


akoeplinger:
An important topic that came up during the meeting is why this needs to be a new API instead of using/adapting the existing stuff in `Environment.OSVersion`. I think the tradeoffs that lead to this decision should be part of the notes since this is something that caused confusion in the initial GH issue.

terrajobst:
That's a great point. I'll add it.

Sent from my phone

On Jun 9, 2015, at 3:05 PM, Alexander K?plinger <notifications@github.com<mailto:notifications@github.com>> wrote:


An important topic that came up during the meeting is why this needs to be a new API instead of using/adapting the existing stuff in Environment.OSVersion. I think the tradeoffs that lead to this decision should be part of the notes since this is something that caused confusion in the initial GH issue.

-

PR #6, opened 2015-07-01
Add 2015-06-30-configuration

terrajobst:


Garciat:
Hello! Pardon my intrusion here. I'd like to point out a couple of things:

**One**
I gathered from the code review video that you were trying to model this (F#):

```fsharp
type Configuration =
    | KeyValue of string * string
    | Section of string * list<Configuration>
```

ADT's can get tricky to implement in OO, as they mix data and behavior concerns into a single type.

**Two**
My vote is for immutable `IConfiguration` (:

**Three**
Concept: use model binding for configuration.

*Rationale:*
One could argue configuration models are in fact static. Components have fixed configuration points. Treating configuration as a dictionary is a bit of an over-generalization.

*Pros:*
* Consumers no longer have to parse string configurations by hand.
* Type-safe, self-validating model.
* Could probably reuse model binding logic in `Microsoft.AspNet.Mvc.ModelBinding`.
* Tooling hints.

*Cons:*
* Reflection-based.
* ...

*Samples:*
```csharp
// user-defined class with bits they care about
public class AppConfigurationModel
{
    // mutable, generic configuration
    public ConcurrentDictionary<String, String> Etc { get; set; }
    
    // delegate sub-configuration to concerning component's types
    public EntityFrameworkConfigurationModel Data { get; set; }
    
    public String BackgroundColor { get; set; }

    // let model binding parse values
    public Int32 SessionLifeMinutes { get; set; }
}

public class Startup
{
    public AppConfigurationModel Configuration { get; }
    
    public Startup(IApplicationEnvironment appEnv)
    {
        Configuration =
            new ConfigurationBuilder(appEnv.ApplicationBasePath)
            .AddJsonFile("config.json")
            .AddEnvironmentVariables()
            .Build<AppConfigurationModel>();
    }
    
    public void ConfigureServices(IServiceCollection services)
    {
        // feed to DI
        services.AddSingleton(appConfiguration);
        // could even do this and let EF pick the config up itself
        services.AddSingleton(appConfiguration.Data);
    }
}
```

We could even leverage attributes to select configuration sources:

```csharp
public class AppConfigurationModel
{
    [FromJson]
    public String DomainName { get; set; }

    [FromEnvironmentVariables]
    public Int32 NumberOfFoos { get; set; }
}
```

terrajobst:
Do you mind replying to the [issue in CoreFX](https://github.com/dotnet/corefx/issues/2199)? I'd like to keep this PR focused on the notes themselves. Thanks!

PR #9, opened 2015-07-28
Add 2015-07-28-long-paths

terrajobst:


2015-07-28-long-paths/README.md
```diff
@@ -0,0 +1,133 @@
+# API Review 2015-07-28
+
+This API review was also recorded and is available on [Google+](https://plus.google.com/events/cu7i66er1ien3s3975o8a5hs8i8).
+
+## Overview
+
+In this API review we discussed how we want address long path limitations due
```
akoeplinger: we want **to** address

2015-07-28-long-paths/README.md
```diff
@@ -0,0 +1,133 @@
...
+[Video](https://plus.google.com/events/cu7i66er1ien3s3975o8a5hs8i8)
+
+### Enabling Long Paths
+
+* .NET Core needs to handle > MAX_PATH (260) paths for Unix
+  - Enable CoreFX (Unix mostly done)
+  - Enable CoreCLR (Assembly & resource loading, not done)
+* Also want to support > MAX_PATH on Windows
+  - Sharing early thinking, will recircle with deeper dives
+  - Many details need uncovered / answered
```
akoeplinger: need **to be** uncovered ?

2015-07-28-long-paths/README.md
```diff
@@ -0,0 +1,133 @@
...
+### Windows Extended Path Syntax
+
+* Paths start with special character sequence
+  - `\\?\` for local paths
+  - `\\?\UNC\` for UNCs (`\\Server\Share`)
+* Allows ~32K length paths
+  - exact path length has to fit when converted to internal NT path
+* Additional characters (trailing spaces, period)
+* Allows legacy device names
+* Does not normalize (`.`, `..``)
```
akoeplinger: remove extraneous backtick at the end

2015-07-28-long-paths/README.md
```diff
@@ -0,0 +1,133 @@
...
+  - Legacy character support (no trailing spaces, etc.)
+  - Extended syntax not given back
+* Intercept before P/Invoke
+  - Append extended if needed
+  - Trim extended on return if added
+
+### Allow Extended Syntax (`\\?\`)
+
+* Don't attempt to normalize
+* Very basic correctness checks (invalid chars, `.`, `..``)
```
akoeplinger: remove extraneous backtick at the end

akoeplinger:
Left a few minor comments, otherwise LGTM :+1:

terrajobst:
Thanks! Fixed.

JeremyKuhne:
LGTM

tmat:
Nice! Yay for Path.PathsAreEquivalent(string firstPath, string secondPath)
Just curious, will it use some kind of file system specific file identity? 

Also, could variants of Path.Combine, Path.GetFileName, Path.GetDirectory, Path.GetExtension that don't check for invalid characters be provided? So that we can work with patterns like Path.GetDirectory("C:\temp\*.cs") == "C:\temp" etc.

terrajobst:
> So that we can work with patterns like  Path.GetDirectory(@"C:\temp\*.cs") == "C:\temp"  etc.
Interesting. I assume you want this API to filter break user input apart, I guess?

tmat:
Re comparing paths: The scenario that might break is when a case-insensitive FS partition is mounted as a subdirectory of case-sensitive FS. Then half of the path needs to be compared case-insensitively and half case-sensitively.

Correct. Roslyn has custom implementations of these Path API for that reason. We'd prefer not having them.

terrajobst:
> The scenario that might break is when a case-insensitive FS partition is mounted as a subdirectory of case-sensitive FS. Then half of the path needs to be compared case-insensitively and half case-sensitively.

Yeah, @jaredpar brought this up in the meeting as well. Hence the proposal to expose an API that gets a canonical representation. In that, all inputs would be using the FS way to get a canonical representation. Afterwards, all comparisons could be done in a case sensitive fashion.

PR #15, opened 2015-09-22
2015-09-22-misc

terrajobst:
@vancem, @kouvel, and @jkotas: please take a look at the notes.

### DIFF
--- a/2015-09-22-misc/README.md
+++ b/2015-09-22-misc/README.md
@@ -1,0 +1,119 @@
+# API Review 2015-09-22
+
+This API review was also recorded and is available on [Google+](https://plus.google.com/events/c4t5gher6o0mhgqnmn53np51nto).
+
+## Overview
+
+In this API review we reviewed the following PRs and proposals:
+
+* [Notes](#telemetrysource) | TelemetrySource
+* [Notes](#3054-assemblyloadcontextloadunmanageddllfrompath) | [#3054](https://github.com/dotnet/corefx/issues/3054) `AssemblyLoadContext.LoadUnmanagedDllFromPath`
+
+## Notes
+
+### TelemetrySource
+
+Status: **Approved with Feedback** |
+[Proposal](https://github.com/dotnet/corefx/tree/master/src/System.Diagnostics.Tracing.Telemetry/src/System/Diagnostics/Tracing) |
+[Video](https://plus.google.com/events/c4t5gher6o0mhgqnmn53np51nto)
+
+* Should be renamed to `DiagnosticSource`
+* `WriteTelemetry(string telemetryName, object parameters)`
+    - should be renamed to `Write`
+    - `telemetryName`  should be `key` (same for `IsEnabled`)
+    - `parameters` should be named `value`
+* `IDisposable` should be implemented correctly
+    - Needs a finalizer
+    - Needs `Dispose(bool isDiposing)` should be virtual
+    - `Dispose()` shouldn't be virtual
+
+#### Details
+
+`TelemetrySource` is a new API that has been added to the CoreFX framework. It
+lives in the DLL `System.Diagnostics.Tracing.Telemetry`.
+
+We would have preferred to not have this class, and instead simply use
+EventSource. However there is a desire for 'Rich Telemetry' where the consumer
+of the telemetry is IN PROCESS but gets RICH data (basically as detailed as what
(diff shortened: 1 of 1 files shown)


terrajobst:
Our friend @praeclarum [reminded us](https://twitter.com/praeclarum/status/646380993100673024) to pay attention to the QA panel.

questions

Here is my take on these questions -- they are all about `TelemetrySource`. @vancem, please feel free to jump in and correct me.

### The hierarchy seems confusing

> Listener is a subclass of `Source`? But you are expected to subclass `Source`?

Producers of events (which are the majority of the users in this system) are not expected to create any derived types. They will simply call the APIs like so:

```C#
TelemetrySource.DefaultSource.WriteTelemetry("IntPayload", 5);
```

Consumers of events (of which we have a very limited number) can simply call the APIs on `TelemetryListener` or create a derived type -- whatever is easier.

Testing that events got produced correctly would derive from `TelemetryListener` so that they mock out the implementation.

### Isn't 1 object allocation "close to nothing"?

That argument was made during the meeting as well. In general the expectation is that many event producers get away with passing in existing objects, such as `HttpContext`. Before new objects are being constructed, they can check with is `IsEnabled` whether it's worth it.

### Shouldn't AllListeners be an ObservableCollection? Replaying history seems like a dangerous hack on IObservable.

Can you elaborate?

### What is the semantic difference between Source.DefaultSource and Listener.DefaultListener? When do I use which?

My understanding is that producers are expected to use `Source.DefaultSource` while consumers would use `Listener.DefaultListener`.

### "Telemetry" isn't the problem - it's the "Source" and "Listener" that are confusing.

> The problem with the names comes from a lack of understanding of roles.
> Why two classes? Which exact role does each fulfill?

That feedback was brought up as well; it seems there is a belief that it helps with separating the concerns.

praeclarum:
Thanks for posting these!

## ObservableCollection

An `IObservable` is not a good way to store a dynamic list. For one, there should not be a history associated with it. Subscribers expect a stream of data objects - not "pieces of an object".

As a concrete example of why this is bad: How does a subscriber note the **removal** of a listener? An IObservable can't do this.

There are two solutions, one good, one bad:

1. `ObservableCollection<Listener>` is designed for exactly this scenario - holding a list and notifying clients of changes.
1. `IObservable<IReadOnlyCollection<Listener>>` is the alternative if you desperately want to support IObservable. Good because it broadcasts state changes - removals are noted.

## Hierarchy

The API allows:

```
source1 : Source = ...
listen1 : Listener = ...
listen2 : Listener = ...

source1.Subscribe (listen1); // Cool
listen1.Subscribe (listen2); // Really, listener chaining is allowed?
```

Just seems like a very convoluted design for a pretty simple/standard pattern.

I don't see why inheritance is needed at all - sources and listeners are different concepts and don't need to be related through a base class.

--

Anyway, just my 2 cents since I probably won't be using this API - not useful enough/designed to be a logging solution.

vancem:
@praeclarum 
The main scenario for 'AllListeners' is to give client (e.g. a tool) the ability to subscribe to the listener.   Thus you are not actually interested in the list, but rather you simply want an opportunity to subscribe.   

Notice that we don't care about removals, all we want is an event for each TelemetryListener (past or present), so that we can subscribe.   

Moreover, we are really trying to keep the dependencies in this code minimal.  ObservableCollection is not sufficiently primitive.

praeclarum:
Fascinating. Just seems odd to build such a limited API with such a restricted use case. You may be building it for tools under some strange auspice, but humans will be looking through this code, wondering why this is public, and then trying to decipher its strange design. So I guess that's 3 cents, sorry. :-)

vancem:
I suppose whether an API is limited or not is a matter of design.   Adding too much is typically worse than adding too little (because you can't every take back exposed APIs).   IEnumerable as by design very impoverished, and by design so was IObservable, it is part of what makes them powerful (lots of things fit that pattern).  

One of the most important properties of this design is that it has very few dependencies (basically it depends on System.Runtime and Threading (only for Interlocks), and to keep it simple/light.   

If indeed we made a mistake by not making AllListeners an ObservableCollection, then frankly it could be implemented 'on top', if so desired.    Time will tell.

praeclarum:
I agree with you, time will tell. In fact there's no reason to make any critical decisions since only time will tell.

PR #17, opened 2015-10-09
Add 2015-10-09-portable-pdb

terrajobst:


2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
++   public DebugMetadataHeader DebugMetadataHeader { get; }
++   public DocumentHandleCollection Documents { get; }
++   public ImportScopeCollection ImportScopes { get; }
++   public LocalConstantHandleCollection LocalConstants { get; }
++   public LocalScopeHandleCollection LocalScopes { get; }
++   public LocalVariableHandleCollection LocalVariables { get; }
++   public MethodBodyHandleCollection MethodBodies { get; }
++   public CustomDebugInformation GetCustomDebugInformation(CustomDebugInformationHandle handle);
++   public CustomDebugInformationHandleCollection GetCustomDebugInformation(EntityHandle handle);
+    ^ Immo Landwerth: Needs to be plural
```
nguerrera: During the review, I proposed Info/Infos, but now I think Information for both is OK, which is current state of PR. We spell out XxxReference, XxxDefintiion, XxxSpecification, so XxxInformation fits (although XxxInfo fits with reflection).  "Informations" (or really even Infos) is not correct, so let's leave this.

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
+  public struct MetadataStringComparer {
++   public bool Equals(DocumentNameBlobHandle handle, string value);
++   public bool Equals(DocumentNameBlobHandle handle, string value, bool ignoreCase);
++   public bool Equals(NamespaceDefinitionHandle handle, string value, bool ignoreCase);
++   public bool Equals(StringHandle handle, string value, bool ignoreCase);
++   public bool StartsWith(StringHandle handle, string value, bool ignoreCase);
+  }
++ [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+  public struct MethodBody {
+    ^ Immo Landwerth: MethodDebugInfo
```
nguerrera: Let's go with MethodDebugInfomation, see comment on CustomDebugInformation.

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
++   public StandaloneSignatureHandle LocalSignature { get; }
++   public BlobHandle SequencePoints { get; }
++   public MethodDefinitionHandle GetStateMachineKickoffMethod();
+    ^ Immo Landwerth: Seems a bit too specific for a generic MD reader.
+  }
++ [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+  public struct MethodBodyHandle : IEquatable<MethodBodyHandle> {
+    ^ Immo Landwerth: MethodDebugInfoHandle
```
nguerrera: MethodDebugInformationHandle (see related comments)

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
++   public static bool operator ==(MethodBodyHandle left, MethodBodyHandle right);
++   public static explicit operator MethodBodyHandle (EntityHandle handle);
++   public static explicit operator MethodBodyHandle (Handle handle);
++   public static implicit operator Handle (MethodBodyHandle handle);
++   public static implicit operator EntityHandle (MethodBodyHandle handle);
++   public static bool operator !=(MethodBodyHandle left, MethodBodyHandle right);
+  }
++ [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+  public struct MethodBodyHandleCollection : IEnumerable, IEnumerable<MethodBodyHandle>, IReadOnlyCollection<MethodBodyHandle> {
+    ^ Immo Landwerth: MethodDebugInfoHandleCollection
```
nguerrera: MethodDebugInformationHandleCollection (see related comments)

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
++   AliasNamespace = 7,
++   AliasType = 9,
++   ImportAssemblyNamespace = 2,
++   ImportAssemblyReferenceAlias = 5,
++   ImportNamespace = 1,
++   ImportType = 3,
++   ImportXmlNamespace = 4,
+  }
++ [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+  public struct ImportsBlobReader : IDisposable, IEnumerator, IEnumerator<ImportDefinition> {
```
nguerrera: Same comment here as SequencePointBlobReader, should it be ImportDefinitionCollection? Enumerators without corresponding Enumerables don't work with foreach, linq....

2015-10-09-portable-pdb/README.md
```diff
@@ -0,0 +1,28 @@
...
+[Video](https://plus.google.com/events/cc1vin923d7nuddasqkd487b8fk)
+
+* Will existing readers trip over due to new tables?
+    - We believe the answer is no, but **@TMat** should clarify
+* Should the functionality live in a separate assembly?
+    - We concluded that it shouldn't be split -- it shares a bunch of infrastructure
+* Some very generic names, especially Document
+    - We should qualify them, e.g. `SourceDocument`
+* Should `DebugMetadataHeader` be collapsed and expose `EntryPoint` as a method
+  on `MetadataReader`, e.g. `GetDebugEntryPoint()` or something?
```
nguerrera: It now has an extra ID property, so this is OK as is.

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
+```diff
+---D:\Temp\System.Reflection.Metadata.dll
++++D:\Src\corefx\bin\Windows_NT.AnyCPU.Debug\System.Reflection.Metadata\System.Reflection.Metadata.dll
+ namespace System.Reflection.Metadata {
+  [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+  public struct BlobReader {
++   public BlobHandle ReadBlobHandle();
++   public object ReadConstant(ConstantTypeCode typeCode);
++   public DateTime ReadDateTime();
+    ^ Immo Landwerth: Where is that used?
```
tmat: Local constants are stored in PDBs. VB allows date time constants.

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
++   public decimal ReadDecimal();
+  }
++ [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+  public struct CustomDebugInformation {
+    ^ Immo Landwerth: CustomDebugInfo
```
tmat: I was thinking about that but then we have -Specification (vs. -Spec), -Implemenation (vs. -Impl) and I renamed MethodBody to MethodDebugInformation. 

It seems more consistent to use full names.

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
+    public struct Enumerator : IDisposable, IEnumerator, IEnumerator<CustomDebugInformationHandle> {
++     public CustomDebugInformationHandle Current { get; }
++     object System.Collections.IEnumerator.Current { get; }
++     public bool MoveNext();
++     void System.Collections.IEnumerator.Reset();
++     void System.IDisposable.Dispose();
+    }
+  }
++ public sealed class DebugMetadataHeader {
+    ^ Immo Landwerth: What does this do?
```
tmat: Represents the header of #Pdb stream. See spec: https://github.com/dotnet/roslyn/blob/master/docs/specs/PortablePdb-Metadata.md

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
++   public readonly MethodDefinitionHandle EntryPoint;
+  }
++ [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+  public struct Document {
+    ^ Immo Landwerth: Very generic SourceDocument.
```
tmat: Will rename.

2015-10-09-portable-pdb/README.md
```diff
@@ -0,0 +1,28 @@
...
+    - It seems the underlying format doesn't have accordances to grow anyways
```
nguerrera: typo: s/accordances/affordances/

2015-10-09-portable-pdb/APIs.md
tmat: I don't think it should be a collection. It's not random access. Just sequential. And the reader is an enumerator. We can potentially add a corresponding enumerable.

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
++   public bool Equals(StringHandle handle, string value, bool ignoreCase);
++   public bool StartsWith(StringHandle handle, string value, bool ignoreCase);
+  }
++ [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+  public struct MethodBody {
+    ^ Immo Landwerth: MethodDebugInfo
++   public StandaloneSignatureHandle LocalSignature { get; }
++   public BlobHandle SequencePoints { get; }
++   public MethodDefinitionHandle GetStateMachineKickoffMethod();
+    ^ Immo Landwerth: Seems a bit too specific for a generic MD reader.
```
tmat: We have a dedicated table to encode mapping of state machine kick off methods to move-next methods. This is used by async and iterator methods. I moved all async stuff that was too specific for VS debugger to a custom debug info. But this one is very straightforward info. There are even custom attributes in NetFX expressing the other direction of the relation (from kickoff method to the state machine type and effectively he MoveNext method).

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
++   public readonly ushort EndColumn;
++   public readonly ushort StartColumn;
++   public bool IsHidden { get; }
++   public override bool Equals(object obj);
++   public bool Equals(SequencePoint other);
++   public override int GetHashCode();
+  }
++ [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+  public struct SequencePointBlobReader : IDisposable, IEnumerator, IEnumerator<SequencePoint> {
+    ^ Immo Landwerth: Should this be SequencePointCollection?
```
tmat: No, it's sequential enumerator, not random access.

2015-10-09-portable-pdb/APIs.md
nguerrera: Makes sense.

2015-10-09-portable-pdb/APIs.md
nguerrera: Collection does not imply random access in BCL. We actually have tons of `XxxCollection : IReadOnlyCollection<Xxx>` in S.R.M...

nguerrera: Guideline is also not to have IEnumerator classes not suffixed with Enumerator (nor Enumerable not suffixed with collection).

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
++     public MethodBodyHandle Current { get; }
++     object System.Collections.IEnumerator.Current { get; }
++     public bool MoveNext();
++     void System.Collections.IEnumerator.Reset();
++     void System.IDisposable.Dispose();
+    }
+  }
++ [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+  public struct SequencePoint : IEquatable<SequencePoint> {
+    ^ Immo Landwerth: The fields should be properties.
```
tmat: Fixing.

2015-10-09-portable-pdb/APIs.md
```diff
@@ -0,0 +1,405 @@
...
++   public readonly int EndLine;
++   public const int HiddenLine = 16707566;
+    ^ Immo Landwerth: Should this be public?
```
tmat: Yes. It's a well-known constant.

2015-10-09-portable-pdb/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Portable PDB
+
+Status: **Approved with Feedback** |
+[APIs](APIs.md) |
+[Spec](https://github.com/dotnet/roslyn/blob/master/docs/specs/PortablePdb-Metadata.md) |
+[Video](https://plus.google.com/events/cc1vin923d7nuddasqkd487b8fk)
+
+* Will existing readers trip over due to new tables?
+    - We believe the answer is no, but **@TMat** should clarify
```
tmat: If the debug info is embedded into the same metadata as type-system metadata then unlikely. They would need to be pretty smart and even then if they add a method in the middle of a MethodDef table, for example, they would eed to patch all references to method defs that follow. It can't be done without understanding the content (or at least schema) of the new tables. 

Otherwise (default) it's two distinct streams/files, the type system metadata is unchanged and it's not applicable to the PDB stream.

2015-10-09-portable-pdb/README.md
```diff
@@ -0,0 +1,28 @@
...
+* Should the functionality live in a separate assembly?
+    - We concluded that it shouldn't be split -- it shares a bunch of infrastructure
+* Some very generic names, especially Document
```
tmat: We already have Parameter, Constant, Handle, etc. Adding Source to Document and DocumentName results in pretty mouthful names like SourceDocumentNameBlobHandle.

2015-10-09-portable-pdb/README.md
```diff
@@ -0,0 +1,28 @@
...
+    - We should qualify them, e.g. `SourceDocument`
+* Should `DebugMetadataHeader` be collapsed and expose `EntryPoint` as a method
```
tmat: No, there is actually now another property: Id on the header. Future versions might add more.

2015-10-09-portable-pdb/README.md
nguerrera: +1 We spent a bunch of time on this in the v1 review and concluded that given all of the concat'ing of type names, we should start as short as possible, so we should stay consistent with that.

2015-10-09-portable-pdb/APIs.md
terrajobst: I'm fine with this.

2015-10-09-portable-pdb/APIs.md
terrajobst: Sounds good.

2015-10-09-portable-pdb/APIs.md
terrajobst: Personally, I could live with the "Reader" suffix as it's quite common in binary parsers and already in use by this library. The only oddity is implementing IEnumerable, which to me implies that the usage is basically identical to collections, which is why I'm siding with @nguerrera.

2015-10-09-portable-pdb/APIs.md
terrajobst: Only `DateTime`? What about `DateTimeOffset`?

2015-10-09-portable-pdb/APIs.md
tmat: Yeah, Nick persuaded me as well.

2015-10-09-portable-pdb/APIs.md
terrajobst: Seems inline with that @nguerrera said above, so I'm fine with that.

2015-10-09-portable-pdb/APIs.md
tmat: Just DataTime:

```VB
 Sub Main()
        Const c = #10-9-2015#
        Dim d = #10-9-2015#
    End Sub
```
Emits (for the variable):
DateTime dateTime = new DateTime(635799456000000000L);

2015-10-09-portable-pdb/APIs.md
terrajobst: So you're saying because there is a metadata table it's a first class concept? I buy that.

2015-10-09-portable-pdb/APIs.md
terrajobst: I'm asking because you also exposed a property that indicates the same. Are consumers expected to compare the line against the constant or to check the property?

2015-10-09-portable-pdb/README.md
terrajobst: But the format is already encoded into streams, right?

2015-10-09-portable-pdb/README.md
terrajobst: @KrzysztofCwalina points was that you could strip the qualifier from other usages, but I tend to concur with @tmat.

2015-10-09-portable-pdb/README.md
terrajobst: Sounds good.

terrajobst:
Updated the notes.

nguerrera:
tagging @weshaggard

terrajobst:
So are we happy with the notes?

nguerrera:
LGTM

terrajobst:
Thanks!

PR #23, opened 2016-01-13
Add 2016-01-12-simd-and-console

terrajobst:


terrajobst:
@jkotas what are your thoughts on exposing these APIs:

```diff
 namespace System.Numerics {
     public struct Vector<T>
     {
+        public Vector(void* dataPointer);
+        public void CopyTo(void* destination);
     }
 }
```

The behavior of the constructor is to copy the `Vector<T>.Length * sizeof(T)` bytes into the underlying SIMD register. `CopyTo` would copy the register into the destination pointer.

For security/memory safety we've in the passed forced developers to pass in the buffer size. Do you think we should the same here? If so, I assume it would be size of the buffer to read from/write to, as opposed to the number of bytes to read/write.

Thoughts?

jkotas:
I think it make sense to pass the unmanaged source or destination buffer sizes as an argument for extra protection here. The right buffer size is not immediately obvious from the signature and it matches the SDL best practices.


It is unfortunate that we are compensating for C# limitations by suboptimal APIs like this one. One would actually want to write `var vector = *(Vector<T> *)dataPointer;`. It is possible in IL, but C# does not allow it. Since it is going to take a while until - if ever - similar C# limitations are addressed, I am wondering whether it would make sense to create a library with unsafe helpers that allow working around these limitations when folks need it. It may be better than adding suboptimal unsafe APIs here and there.

```
public static class Unsafe
{
      [MethodImpl(MethodImplOptions.AggressiveInlining)]
      public static T Read<T>(void * p)
      {
          ldarg.0
          ldobj !!T
          ret
      }

      [MethodImpl(MethodImplOptions.AggressiveInlining)]
      public static void Write<T>(void * p, T value)
      {
          ldarg.0
          ldarg.1
          stobj !!T
          ret
      }
}
```

There are other useful unsafe helpers, in addition to two above ... I have written them countless of times for my various experiments.

terrajobst:
> I think it makes sense to pass the unmanaged source or destination buffer sizes as an argument for extra protection here. The right buffer size is not immediately obvious from the signature. Passing explicit buffer size matches the SDL best practices.

Sounds good. @mellinoe could you make that change and update this issue accordingly?

> It is unfortunate that we are compensating for C# limitations by suboptimal APIs like this one.

So you're saying that folks would use the API like this:

```C#
void* data = GetData();
Vector<float> data = Unsafe.Read<Vector<float>>(data);
```

Let's say were to do this. I'd we wouldn't need to pass in the right buffer size because callers would expect to be `sizeof(Vector<float>)`?

terrajobst:
Both casts are unsafe because you're basically doing a reinterpret cast from or to a `void*`. Also, I'm not sure we'd want extension methods on an (essentially) unconstrained `T` because that would mean they show up on every value type. Besides, the extension method design wouldn't work because your `UnsafeCast<T>` would return a pointer to a value type that is passed on the stack. By the time the method returns, the stack frame is popped and thus is likely causing problems when used later.

jkotas:
> I assume we wouldn't need to pass in the right buffer size because callers would expect it to be  sizeof(Vector<float>) ?

Correct. I think about these methods as a way to express `*(Vector<float> *)dataPointer` - that does not have any size checks either.

BTW: Another good addition to the `Unsafe` class would be sizeof(T) equivalent that would allow folks to write the proper buffer size checks when they need to.

```
      [MethodImpl(MethodImplOptions.AggressiveInlining)]
      public static int SizeOf<T>()
      {
          sizeof !!T
          ret
      }
```

benaadams:
They'd both show as Unsafe as one you'd have to call through the class `Unsafe`.

But yeah; all your other points kind of make it moot anyway...

mellinoe:
Thanks for the discussion, some good points brought up here. A couple of counter-points / points for further discussion:

Like you mentioned, @jkotas , the main purpose of these overloads is to workaround the fact that we cannot do this in C#:
````CSharp
 var value = *(Vector<float> *)dataPointer;
*(Vector<float> *)dataPointer = value;
````
I'm a bit concerned that making users specify their buffer size and then doing argument validation will slow down these methods a bit too much. After all, they're supposed to be the replacements for the totally-unsafe examples above, with which you can definitely shoot yourself in the foot with if you dereference or write to an inappropriate pointer location.

A counter-example in the BCL is `Marshal`, has some overloads that don't require you pass a buffer length. `Marshal.WriteInt32(IntPtr, Int32)` is one such example. To be fair, `Marshal` is a bit of an atypical class.

@CarolEidt How much of an impact would it be on the code-gen for these overloads if we needed to do argument validation, i.e. verify that the length passed in is at least `Vector<T>.Count`?

Another thing that we decided was that we would remove the `offset` or `index` parameter from the overloads, because you can just pass a different pointer value instead of passing a "base pointer" and an offset. In that case, it would be a little awkward to have to pass in your "buffer length" because you may pass a pointer to the middle of your buffer. What you'd really need to pass would be the length remaining in the buffer starting from your pointer. And the only thing we'd really check is that the length remaining is greater than or equal to `Vector<T>.Count`, because the size of the data written is always that much.

The ideas about an `Unsafe` class are also definitely interesting. Even with them, i think that we would still like to have these overloads available on `Vector<T>` because we want them to be recognized by the JIT so that it can make particular/special decisions regarding the code-gen for `Vector<T>`. I suppose the JIT could also recognize `Unsafe`, but it would probably not mesh as well with the existing intrinsic recognitions.

**Comment by jkotas** [comment]
>  I suppose the JIT could also recognize  Unsafe , but it would probably not mesh as well with the existing intrinsic recognitions

The Unsafe class should just work as expected today, without any extra work in the JIT. The JIT does not need to recognize it in any special way. It is the beaty of it.

Forgot to mention in my earlier comment: The methods in System.Numerics namespace are 100% safe There is not a single method that takes a pointer today. It was also one of the reasons why I have leaned towards passing the extra buffer sizes here - to make people think twice about the buffer sizes.

**Comment by benaadams** [comment]
You could take the buffer size as a param; then not check the values - best of both :stuck_out_tongue_winking_eye:

**Comment by benaadams** [comment]
Precedent for passing in the remaining lengths would be [Buffer.MemoryCopy](https://github.com/dotnet/coreclr/blob/bc146608854d1db9cdbcc0b08029a87754e12b49/src/mscorlib/src/System/Buffer.cs#L522)
````
public static unsafe void MemoryCopy(void* source, void* destination, long destinationSizeInBytes, long sourceBytesToCopy)
```

However on the against; `MemoryCopy` would be a one shot deal whereas the `Vector` functions would likely be used in a tight loop and if used correctly would incur an additional decrement to bufferSize and "bounds" check comparision.

benaadams:
Also related `PtrUtils` in `System.Slices` in corefx
https://github.com/dotnet/corefxlab/blob/67d6fa0556d0835c2d779707904d7474ef9da81e/src/System.Slices/src/System/PtrUtils.cs

mellinoe:
> The Unsafe class should just work as expected today, without any extra work in the JIT. The JIT does not need to recognize it in any special way. It is the beaty of it.

I know that the Unsafe methods discussed above would work in the general case, but I was more referring to the special handling of vector intrinsics by RyuJIT, which I guess would have a harder time recognizing the usage of `Unsafe.Write<Vector<float>>(dest, vec)` than `vector.CopyTo(dest)` because of the way the intrinsic recognition is handled.

> if used correctly would incur an additional decrement to bufferSize and "bounds" check comparison for each loop iteration.

That's my main concern. From what I understand, the current overloads can often translate very closely to a single load or store instruction, and can often have the array bounds checks eliminated. Having to do a validity check every time would probably make the story a bit different. Perhaps @CarolEidt knows how much it would affect the code-gen.

CarolEidt:
I don't have a good understand of what the Unsafe class would translate to, in terms of IL. Currently we do "bounds checking" on the Vector<T> index methods using a mechanism that correlates rather closely with array bounds checking. This makes it easier for the JIT optimizations to hoist or eliminate these checks (which, AFAIK is not yet fully supported). Doing something materially different would make it more difficult to eliminate these checks, which would probably add an order of magnitude overhead to what would otherwise be a single load or store.

jkotas:
> I was more referring to the special handling of vector intrinsics by RyuJIT, which I guess would have a harder time recognizing the usage of  Unsafe.Write<Vector<float>>(dest, vec)

There is *no* special handling required for Unsafe.Read/Write in the JIT to make it work perfectly for Vector<T>. I have done a quick test:

```
static void add<T>(void * a, void * b, void * result) where T:struct
{
    Unsafe.Write<Vector<T>>(result, Unsafe.Read<Vector<T>>(a) + Unsafe.Read<Vector<T>>(b));
}
```

This is the code for add<int> with current RyuJIT:
```
repro!ConsoleApplication.Program.add[[System.Int32, mscorlib]](Void*, Void*, Void*):
movups  xmm0,xmmword ptr [rcx]
movups  xmm1,xmmword ptr [rdx]
paddd   xmm0,xmm1
movups  xmmword ptr [r8],xmm0
ret
```

I have shared the IL at https://gist.github.com/jkotas/ce440d1a2fc869581535 if you would like to try it yourself.

omariom:
Unsafe class got my vote :+1:

CarolEidt:
The Unsafe helpers seem like a great addition, and should be considered in their own right, but there would still be a discoverability issue for developers. In the interest of discoverability + performance (which I think that most SIMD developers will care about), we would be better off having the simple void* overloads to avoid requiring the JIT to add checks that would be unlikely to be optimized away, even in a loop, without JIT changes.

mellinoe:
> I have shared the IL at https://gist.github.com/jkotas/ce440d1a2fc869581535 if you would like to try it yourself.

Thanks, @jkotas, i'm able to see the same thing on my desktop here. I'd like to try it on my AVX2 laptop as well, but I left it at home today. It definitely seems that the JIT is doing exactly the right thing here.

My concerns are the same as Carol's: adding the size validation will be too expensive to use in a tight loop if we aren't able to optimize them out. I agree that `Unsafe` would be a great addition, but I still would like to have a good option on `Vector<T>` itself so that it would be easier to discover and would fit in more nicely with the rest of the API.

If we did have an `Unsafe` class, would we also expect that the buffer lengths need to be passed in explicitly as well (i.e. `Unsafe.Write(void*, T, int)`)? Or is the main concern that `Vector<T>` is a 100% safe API right now, and we ought to be extra careful if we have a one-off unsafe API that people could shoot themselves in the foot with? Whereas on a class called `Unsafe` where you probably expect Bad Things to happen if you aren't careful? I can definitely understand that argument, I'm just trying to rationalize it with the fact that we'd ideally like the API's to be as fast as possible, basically functionally identical to the `Unsafe` ones above.

jkotas:
I will just reiterate my earlier points:

The Unsafe methods are a direct equivalent of `*(Vector<float> *)dataPointer` that does not have any size checks either (without generics). Not requiring buffer checks in these super low level cases should be acceptable. It is a good thing that the Unsafe methods would not be easily discoverable, and that the syntax to use them is more verbose.

The constructors that takes void* does not look as super low level to me. The SDL best practices are to have explicit buffer sizes in APIs like this. It is not a good thing that the unsafe constructor is easily discoverable. The fact that everything else in this namespace is safe does not help either.

There was a similar very long discussion about this when the Buffer.MemoryCopy API was designed - whether it should take redundant source buffer size argument just an extra security check. The bottom line of this discussion was that the security aspect should win.

If other security experts are ok with the void* argument and no buffer size here, I do not have a problem with being overruled.

@morganbr Do you have opinion on this one?

davidfowl:
Reminds me of this http://mishadoff.com/blog/java-magic-part-4-sun-dot-misc-dot-unsafe/. It's also super undocumented in java land.

terrajobst:
@davidfowl are you saying that's evidence that we shouldn't do the `Unsafe` class approach?

FWIW, .NET always had unsafe capabilities, especially around interop. Customers are somewhat used to the idea that, for instance, we've special operations on the `Marshal` class and our `Buffer` classes (to which we recently added `MemoryCopy`).

nguerrera:
I also strongly favour the general Unsafe helpers over adding one-off unsafe surface to Vector. By developing simple, orthogonal unsafe features, we can limit the spread of redundant unsafe API to domain-specific libraries.

morganbr:
My thoughts on this are that we should try to optimize for everybody's memory safety. I have some more general ideas that I'm still formulating, but for this case:

1. I think it's absolutely critical that at some point the buffer length gets checked. Exploits of unchecked lengths are far too common to force users to try to get it right (especially for something where they could pretty easily get the size wrong). That said, even taking it as a parameter isn't great since people make mistakes on that fairly regularly, but it's my minimum bar.
2. I'm not entirely opposed to the Unsafe<T> idea, but I'd rather see it restricted only to value types with no internal pointers (this might require some runtime or JIT work). If it could blit arbitrary structs, I think the odds of having generic code accidentally write arbitrary pointers is concerning. 
3. I'd like to understand the scenario better -- is this a void* because somebody pinned a managed byte[] and wants a Vector out of the middle of it? Is it a memory mapped file? A value from native code? Based on the scenario, is there some safe code construct we could build around it? For example, would it be possible to build a Span over the managed byte[] and then have Unsafe take a Span and do proper bounds checking?

jkotas:
I think that the Unsafe class should have no artificial restrictions. The current unsafe context in C# has artificial restrictions, and that's why we are having this discussion - the artificial restrictions just get into way.

If you are using the Unsafe class, you have to know what you are doing and be very careful. If you do not know what you are doing, you will get it wrong - the artificial restrictions won't save you. The artificial restriction just get into a way for folks who know what they are doing. E.g. CoreRT is full of [shenanigans](https://github.com/dotnet/corert/blob/master/src/Runtime.Base/src/System/Runtime/RuntimeExports.cs#L162) peppered with IL rewriting to compensate for lack of intentionally-designed unrestricted unsafe access.

morganbr:
There may be a place for unrestricted unsafe actions, but I don't think this feature calls for it -- converting a byte array to a struct Vector is a completely safe/verifiable action if there are bounds checks and no GC pointers.

mikedn:
> Based on the scenario, is there some safe code construct we could build around it? For example, would it be possible to build a Span over the managed byte[] and then have Unsafe take a Span and do proper bounds checking?

Yep, you can do that and you can also build a span from a void* and a length:

```C#
IntPtr ptr = Marshal.AllocCoTaskMem(256);
Span<float> spf = new Span<float>((void*)ptr, 256 / sizeof(float));
Span<Vector<float>> spvf = spf.Cast<float, Vector<float>>();
Vector<float> vf = spvf[0];
```

This makes the addition of an unsafe and bounds checking `Vector<T>` constructor questionable since spans should fulfill that need.

I'm for Unsafe class. Not only that it is a more general solution but it also raises the visibility of this type of code in reviews, investigations etc.

benaadams:
> I'd like to understand the scenario better -- is this a void* because somebody pinned a managed byte[] and wants a Vector out of the middle of it? Is it a memory mapped file? A value from native code?

A classic scenario for unsafe code is bounds check elimination on src+dst with datatype widening and loop unrolling. Given the below code, each Vector `.ctor` and `CopyTo` does a bounds check as it is considered in isolation by the internals of the calls and they are being used in a tight loop.

Further down for `long*`/`short*`/`int*`/`byte*` based cast copy a `fixed` block has been opened to make this possible. It would be more efficient to open the `fixed` block at the top of the first `while` loop and pass pointers into the `.ctor` and `CopyTo` allowing these already done checks to be eliminated.

```csharp
public unsafe static void VectorizedCopy(
    this byte[] src, 
    int srcOffset, 
    byte[] dst, 
    int dstOffset, 
    int count
)
{
  if (Vector.IsHardwareAccelerated) {

if (src == null) throw new ArgumentNullException(nameof(src));
if (dst == null) throw new ArgumentNullException(nameof(dst));
if (count < 0 || srcOffset < 0 || dstOffset < 0) throw new ArgumentOutOfRangeException(nameof(count));
if (count == 0) return;
if (srcOffset + count > src.Length) throw new ArgumentException(nameof(src));
if (dstOffset + count > dst.Length) throw new ArgumentException(nameof(dst));

while (count >= _vectorSpan4)
{
    new Vector<byte>(src, srcOffset).CopyTo(dst, dstOffset);
    new Vector<byte>(src, srcOffset + _vectorSpan).CopyTo(dst, dstOffset + _vectorSpan);
    new Vector<byte>(src, srcOffset + _vectorSpan2).CopyTo(dst, dstOffset + _vectorSpan2);
    new Vector<byte>(src, srcOffset + _vectorSpan3).CopyTo(dst, dstOffset + _vectorSpan3);
    if (count == _vectorSpan4) return;
    count -= _vectorSpan4;
    srcOffset += _vectorSpan4;
    dstOffset += _vectorSpan4;
}
if (count >= _vectorSpan2)
{
    new Vector<byte>(src, srcOffset).CopyTo(dst, dstOffset);
    new Vector<byte>(src, srcOffset + _vectorSpan).CopyTo(dst, dstOffset + _vectorSpan);
    if (count == _vectorSpan2) return;
    count -= _vectorSpan2;
    srcOffset += _vectorSpan2;
    dstOffset += _vectorSpan2;
}
if (count >= _vectorSpan)
{
    new Vector<byte>(src, srcOffset).CopyTo(dst, dstOffset);
    if (count == _vectorSpan) return;
    count -= _vectorSpan;
    srcOffset += _vectorSpan;
    dstOffset += _vectorSpan;
}
if (count > 0)
{
    fixed (byte* srcOrigin = src)
    fixed (byte* dstOrigin = dst)
    {
        var pSrc = srcOrigin + srcOffset;
        var dSrc = dstOrigin + dstOffset;
```
As this is a straight forward copy the bound check elimination could arguably be taken care of (hidden) by `MemoryCopy` in a single call except it doesn't currently have access to types wider than long. Also since its managed code it would be effected by the same restrictions.

However that doesn't cater for more complex operations. For example if instead of a straight copy it was also performing a transformation, for example lowercasing ascii:

```csharp
var A = new Vector<byte>(65); // A
var Z = new Vector<byte>(90); // Z
var lowerCaseBit = new Vector<byte>(0x20); // 0010 0000
...
var input = new Vector<byte>(src, srcOffset);
Vector.ConditionalSelect(
    Vector.BitwiseAnd(
        Vector.GreaterThanOrEqual(input, A),
        Vector.LessThanOrEqual(input, Z)
    ),
    Vector.BitwiseOr(lowerCaseBit, input),
    input)
   .CopyTo(dst, dstOffset);
```
Applying extra processing means a framework call like `MemoryCopy` can no longer cater for the bounds check elimination. This transform step would likely be more complex either operating over a Structure of Arrays of `Vector<T>`; or over an array of `Vector3/4`s dependent on the application with operations applied via `Quaternion` and `Matrix4x4`.

Native interop use cases which also have this quality are processing XAudio2 sound buffers in UWP apps, prepping data for DirectX buffers which may be kinematic or physics data updates for the GPU; I'm also sure there is a whole host of applications outside of games, but we have our specialities of thought :grin: 

Our particular use case is for server-side physics calculations for networked games; where after the calculation step we additionally have to then process the data into a lossy compressed byte format for transmission on the wire - rather than interop with DirectX etc.

The Vector types have opened up C# for a whole host of applications that would traditionally be served by C++ (with its greater number of pitfalls); and the pointer cast would only add to this potential.

On the flipside... am happy if the functions aren't not immediately discoverable; with there inhernt dangers of buffer overrun - however aren't you already in that place if you are using unsafe code and pointers?

davidfowl:
> @davidfowl are you saying that's evidence that we shouldn't do the Unsafe class approach?

That's evidence we *should* do it.

benaadams:
Using `Vector` types with strings without an intermderary `char[]` create would also need pointers

omariom:
To use Unsafe in a more safe way CLR and C# need  contains_no_references constraint. And it would be useful not only in the context of memory safety.

nietras:
We do a lot of data processing (e.g. image processing, machine learning etc.) for this we always have data from sources (e.g. cameras) in unmanaged memory, so the proposed functionality with support for pointers would be a tremendous improvement.

Since we would use these inside loops where we have already done bounds checking we really need the methods with no bounds checking, this would incur too much of an overhead. Basically, we would like something as close to the SIMD intrinsics in C++ where you can have load/stores directly on/from pointers. This would no doubt also be useful for other things such as basic foundational memory copying etc.

So for us if these methods would be inlined directly as e.g. _mm256_load_pd/_mm256_store_pd instructions that would be the best case scenario.

However, I think the proposed Vector<T> API is somewhat cumbersome when you have typed pointers, since you have to write Vector<double>(ptr) to load a double type. Also I think static functions would be better since this reads better according to underlying implementations e.g. load/store. 

I think you should, perhaps additionally, add static typed functions on a static non-generic Vector class e.g.
```csharp
public static class Vector
{
    public static Vector<float> Load(float* source);
    public static void Store(Vector<float> source, float* destination);
    public static Vector<double> Load(double* source);
    public static void Store(Vector<double> source, double* destination);
    // And all other types...
    public static Vector<byte> Load(byte* source);
    public static void Store(Vector<byte> source, byte* destination);
}
```

Having Vector<T>.Load/.Store would, though, have the added benefit of perhaps supporting generic code in the future when T is unmanaged if this ever gets supported, see https://github.com/dotnet/coreclr/issues/2322 

So if only one set of APIs would be supported then Vector<T>.Load/.Store, the static non-generic class Vector can be created by users if necessary. In case you want to limit the API surface.

Although, I guess not part of this API review we would also love to have support for converting between types e.g. going from uin8 to int16 is something done in many image processing cases.

nietras:
I guess I should be more clear. I give a +1 for the proposed API:

```csharp
     public struct Vector<T>
     {
+        public Vector(void* dataPointer);
+        public void CopyTo(void* destination);
     }
```
Ctor and CopyTo with pointers only (no size or bound checking) are fine and would be a great addition, although I would prefer Load/Store as naming, but that's probably not in line with normal .NET naming conventions. 

So basically a +1 from me, but please do not try to do all the Span and other stuff it seems overly complicated for something where we really like to be close to the metal. As @benaadams already commented, this funtionality is usually used inside a tight loop that has already been bounds checked.

nietras:
One thing I am, however, wondering about and that is how aligned/unaligned loads/stores would be supported by this? Will the JIT generate code for checking alignment outputting best possible versions or? 

Unaligned access is a lot slower, so LoadAligned, LoadUnaligned, StoreAligned, StoreUnaligned would allow for specific usage which in many cases can give a great speedup e.g. 2x.

nietras:
I didn't read the Unsafe proposal correctly, if the Unsafe static class approach is favored I would suggest perhaps going all in with:
```csharp
public static class Unsafe
{
      [MethodImpl(MethodImplOptions.AggressiveInlining)]
      public static T Load<T>(void * p) // Default load is unaligned

      [MethodImpl(MethodImplOptions.AggressiveInlining)]
      public static void Store<T>(void * p, T value) // Default store is unaligned // Any reason for destination as first argument?  I would prefer destination as last

      // Special cases for using aligned load/stores, 
      // but I presume this would special handling by the JIT
      [MethodImpl(MethodImplOptions.AggressiveInlining)]
      public static T LoadAligned<T>(void * p)

      [MethodImpl(MethodImplOptions.AggressiveInlining)]
      public static void StoreAligned<T>(void * p, T value) // Any reason for destination as first argument? I would prefer destination as last
}
```

Would this work for primitive types too?

Read/Write for me has too much of a mental overlap with file, network, stream etc. I do not think there are any existing BCL APIs that use Read/Write for memory, buffers is there?

jkotas:
`Read/Write` is used for a very similar concepts on `System.Threading.Volatile` and `System.Runtime.InteropServices.Marshal`.

> Would this work for primitive types too?
Yes. I would like it to work for *any* type.

The default should be aligned access because of aligned access is the default everywhere (C#, C++, ...). If unaligned access is interesting, it should have the longer name with `Unaligned` in it.

nietras:
@jkotas you are right, of course. Read/Write then. The more I think about this Unsafe approach the more I like it. It opens up a whole slew of possibilities for working with data in unmanaged memory. In fact, I can't believe I didn't think of this myself. 

Whether its Read/Write, Load/Store, Get/Set, CopyFrom/CopyTo is less important, having this functionality would be awesome! With it we can do what we want, I think, without the unmanaged constraint, although it would improve the type safety of this.

Does alignment hold for SIMD types too if e. g.  the source is a byte* and we Read<Vector<byte>> from it? Just to be sure.

CarolEidt:
@nietras - the JIT currently only assumes alignment for SIMD types on the stack. For reads or writes to non-stack locations it uses unaligned writes. It is my understanding that these incur 0-1 cycle of additional overhead for modern processors if the memory location is actually aligned.

terrajobst:
@morganbr

> I'd like to understand the scenario better -- is this a void* because somebody pinned a managed byte[] and wants a Vector out of the middle of it? Is it a memory mapped file? A value from native code?

All of them. The goal is to expose the processor instructions that loads into a SIMD register from a memory address or stores the SIMD vector into a memory address. The scenarios would encompass anything that deals with data in native memory, which seems to be quite common, considering how many requests we've gotten for this.

@mikedn 

> This makes the addition of an unsafe and bounds checking Vector<T> constructor questionable since spans should fulfill that need.

I think you misunderstand what `Span<T>` and `Vector<T>` are for.

A `Span<T>` is a representation of an arbitrarily sized buffer. The big features are the unification of managed and native memory and representing slices of buffers without having to allocate a new buffer. To make this work, `T` must be a struct that can only contain primitive types or structs of primitive types.

`Vector<T>` isn't arbitrarily sized. While the size isn't known statically it's defined by the size of the SIMD register the CPU offers. Thus, the developer is expected to break down the data into multiples chunks, sized `Vector<T>.Length`. Also, we provide common math operations for `Vector<T>` that the JIT translates into processor specific SIMD instructions. For SIMD, `T` is constrained to the known set of numeric types the .NET platform offers, so it's technically a closed set.

@davidfowl

> That's evidence we *should* do it.

:thumbsup: :-)

terrajobst:
To summarize:

1. It seems we're in agreement that having an `Unsafe` class that offers generic read and write operations.
2. Some folks would prefer still having the proposed members for `Vector<T>`.

My thinking would be:

* Let's do only (1) for now. We can always add the APIs later if this becomes an issue.
* It seems quite valuable to leave `System.Numerics.Vectors` completely safe.
* This also avoids having to design the APIs with buffer sizes as the generic read and write operations are easy to understand to be of `sizeof(T)`. This also follows prior are with APIs we exposed on `Marshal`

@jkotas @morganbr @mellinoe @CarolEidt: any push backs?

CarolEidt:
This seems like quite a reasonable position for now, and good arguments have been made that the less-discoverable nature of the `Unsafe` approach is actually a 'feature'.

mikedn:
> I think you misunderstand what  Span<T>  and  Vector<T>  are for.

Hmm, I don't think so. Maybe my comment was too brief. What I'm saying is that if someone is trying to read `Vector<T>`s from a native buffer but still maintain some level of safety then they can do so by using spans which offers bounds checking. For this reason a `Vector<T>(void *)` constructor or a `Unsafe.Read<T>(void *)` method shouldn't be saddled with bound checks.

terrajobst:
@mikedn 

Ah, I see. Yes, it seems we're saying the same thing. I think the difference between a the constructor and the `Unsafe` class is that it's not clear from the signature how many bytes are read from the `void*`. However, the read function makes it clear that it reads a `T`. So it's more obvious that is has to read `sizeof(T)` bytes.

jkotas:
@jkotas any push backs?

No pushbacks at all. Sounds great to me.

There is a few things to decide for (1) - assembly name, namespace, repo where it lives (corefx vs. corefxlab), etc.

mellinoe:
This seems reasonable to me as well.

I was also going to remark that, if we offered both an `Unsafe` check-less option, and a bounds-checked constructor option, folks would probably naturally use the bounds-checked constructor, assuming it would have performance characteristics that it unfortunately does not. It may be better to force folks who want those performance characteristics to use the `Unsafe` class which would make it more obvious that bounds checks need to be handled manually.

And, to be honest, I am not very familiar with `Span<T>` yet, but @mikedn 's description does seem to indicate that it would directly replace the bounds-checked constructor for some use cases.

Perhaps in the future we will get some language feature that obviates the need for any special handling for this in the library. In that case (assuming something like that ever gets into the language :smile:), having a void* constructor may be more confusing than helpful.

terrajobst:
Sounds good. In that case, I'll update the PR to state that we decided against those members and that we'll come back with a design for the `Unsafe` class.

Thanks for the discussion!

nietras:
In this discussion there was a lot of references to having a way of expressing that a value type only contains value types, this has been filed as an issue by me as https://github.com/dotnet/coreclr/issues/2322

PR #25, opened 2016-01-19
Add 2016-01-19-reflection

terrajobst:


terrajobst:
@pallavit thanks! Addressed with df18868.

PR #27, opened 2016-01-30
Add 2016-01-29-reflection

terrajobst:
@weshaggard @KrzysztofCwalina @pallavit


2016-01-29-reflection/System.Reflection.md
```diff
@@ -0,0 +1,590 @@
...
+---System.Reflection.dll Old
++++System.Reflection.dll New
+ namespace System.Reflection {
+  public sealed class AmbiguousMatchException : Exception {
+    public AmbiguousMatchException();
+    public AmbiguousMatchException(string message);
+    public AmbiguousMatchException(string message, Exception inner);
+  }
+  public abstract class Assembly {
++   public virtual string CodeBase { get; }
```
jaredpar: What does `CodeBase` mean on a UWP application? Won't all `Assembly` instances have the same value?

2016-01-29-reflection/System.Reflection.md
```diff
@@ -0,0 +1,590 @@
...
++   CreateInstance = 512,
++   DeclaredOnly = 2,
++   Default = 0,
++   FlattenHierarchy = 64,
++   GetField = 1024,
++   GetProperty = 4096,
++   IgnoreCase = 1,
++   Instance = 4,
++   InvokeMethod = 256,
++   NonPublic = 32,
```
jaredpar: Private reflection!  I promise not to do too many evil things with it.  :smiling_imp:

bartonjs: @jaredpar Where, clearly, "too many" is defined to be "one more than Jared would do". (Why is there no good smirking emoticon?  You already stole the imp...)

jkotas: The private reflection is in the .NET Core surface already. This flag is not exposing any new fundamental capability.

It is actually very easy to accidently use private reflection in the existing .NET Core surface because of you will get it by default.

The old way of finding methods that we are adding back here is strictly better wrt private reflection because of you to explicitly opt in into it.

2016-01-29-reflection/System.Reflection.md
terrajobst: @pallavit @weshaggard

2016-01-29-reflection/System.Reflection.md
terrajobst: Plus, private *reflection* is usually fine. It's * invocation* what causes the problems :-)

jaredpar: @terrajobst most of the time I don't invoke any methods with private reflection.  I just read and write important field values :smile:

terrajobst: @jaredpar Technically this requires calling a method which is why call this *invocation* as well :tongue:

2016-01-29-reflection/System.Reflection.md
weshaggard: CodeBase\Location wouldn't have any real meaning on UWP, similar to if they were coming from a dynamically generated assembly.  @pallavit do you know what the value actually returns for UWP?

jaredpar: @weshaggard if the assembly is dynamic though the `CodeBase` never has any meaning.  It's a value I can depend on to be worthless 100% of the time.  In non-dynamic assemblies though the value, or lack of value, of this API will change based on where my code is run.  

I'm a bit skeptical of this API being brought back.  In made a lot of sense in the original .NET world where we referenced implementation assemblies and had a single runtime.  In this new world it's making a lot less sense:

- The assembly you reference is very rarely the assembly you run against in .NET Core.  Hence `CodeBase` is often going to be a duplicated value when the user, incorrectly, expects it to be unique.
- The runtime executing the code can change whether or not this value has any meaning. 

What was the motivating scenario for bringing this back?

terrajobst: In many cases, folks use it to determine where to load additional files from. Think of a library that, for instance, includes data files. Folks are doing stuff like:

```c#
Stream GetData()
{
     var path = Path.Combine(GetType().GetTypeInfo().Location, "data.dat");
     return File.OpenRead(path);
}

Comment (jaredpar) [inline 2016-01-29-reflection/System.Reflection.md]: @terrajobst why not `AppContext.BaseDirectory`?

Comment (terrajobst) [inline 2016-01-29-reflection/System.Reflection.md]: That is the replacement. Goal of this effort is to minimize breaks.

File: 2016-01-29-reflection/README.md [hunk 3]
```diff
@@ -0,0 +1,30 @@
...
+
+* Don't expose the notion of `ReflectionOnlyLoad`
+* `GetExecutingAssembly()`
+* Don't expose any `RuntimeXxxHandle` APIs
+* Don't expose `Resolve` methods that resolve tokens to `*Info` classes
+* * Don't expose `ReflectedType`
+* Don't expose `GetCurrentMethod`
+* `GetInterfaceMap()` isn't implemented by .NET Native and would be costly
+    - Use is unclear; can be replaced by calling through the interface
+    - Used by DCS and JSON
```
mconnew: WCF used to have a dependency on GetInterfaceMap but we were able to replace it. See dotnet/WCF#163 for the PR where @roncain replaced usage of GetInterfaceMap with alternate code. DCS and JSONDCS work today without it, is there a scenario which isn't working currently?

terrajobst: Not that we know of; hence we decided not to expose it for now as bringing it to .NET Core would be costly, due to .NET Native.

PR #28, opened 2016-02-03
Add 2016-02-02-ADO_NET

terrajobst:
@saurabh500 @NickCraver @nvivo @FransBouma


NickCraver:
This looks pretty good to me. The one addition I'd make is noting the interface implementation will have to be explicit due to the type forwarding constraints we're working with; it may help whoever happens upon this later.

Thanks for having us today, I learned a few things on the design constraints and thinking which is much appreciated.

terrajobst:
@NickCraver 

> he one addition I'd make is noting the interface implementation will have to be explicit due to the type forwarding constraints we're working with; it may help whoever happens upon this later.

Good point, just added.

PR #29, opened 2016-02-05
Add 2016-02-05-IOperation

terrajobst:


nguerrera:
Adding API-level feedback from my notes that did not already come up in the review or were not captured in the already captured in @terrajobst's notes.

* `IInstanceReferenceExpression` is currently also an `IParameterReferenceExpression` to the hidden 'this' parameter. AFAICT, the hidden 'this' parameter is not otherwise exposed from `IMethodSymbol` and isn't really very useful so we might prefer `IInstanceReferenceExpression` inheriting directly from `IReferenceExpression`. This would resolve another case of a node-type that can be both interior and leaf, which came up as something to avoid across the board.

* Nodes of type `ICase` have `OperationKind.SwitchSection` and nodes of type `ICatch` have `OperationKind.CatchHandler`. These are rare departures from the 1:1 between interface names and operation kinds. We should rename one way or the other to make them match.

* Consider having a single `OperationKind.Branch` for `IBranchStatement` with a separate `BranchKind` enum member with `GoTo`, `Break`, `Exit` and `Continue` values. In general, I'd fined a strict 1:1 between the operation kinds and the leaf-node interfaces cleaner and easier for analyzer authors to internalize as a pattern.

* There are both `ILabeledStatement` and `ILabelStatement` corresponding to differences in C#/VB. `IOperation` should choose one representation or the other and normalize accordingly.

nguerrera:
@terrajobst it's hard to navigate the giant .md for additional feedback. Could you trim it down to just the types that actually have notes?

terrajobst:
I've done this for reflection APIs and doing it manually is a bit of work that it's quite messy due to parentheses and lost context. I should probably add this to API reviewer's diff export.

For now, just search for `^` in the raw file. All comments are preceded by that character.

PR #30, opened 2016-02-11
Add 2016-02-09-ADO_NET

terrajobst:


saurabh500:
LGTM

NickCraver:
Notes the the API review: please don't make the indexer throw on a missing key - I don't see how any consumers would want that. It would make the access of the indexer extremely messy *everywhere* it's consumed. I think if everyone would have to implement practically the same extension method to try/catch and swallow (also inefficient), we've made a bad API call.

terrajobst:
I'd say that we should one of two things:

1. Agree that `null` is a good enough value for non-existing keys.
2. Follow `Dictionary<K,V>` and throw for non-existing keys **and** also offer `TryGetValue` to avoid exceptions.

PR #33, opened 2016-11-06
Add 2016/11-04-SpanOfT

terrajobst:


2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+Instead, we'll introduce an API like `ref T DangerousGetPinnableReference()`.
+Since the API doesn't return a pointer, it doesn't require an `unsafe` context.
+However, the API is inherently unsafe as assigning to the returned value is
+possible but incorrect in case the span has zero-length. Thus, we follow the
+convention and use the term `Dangerous` in the API.
+
+We don't expect user code to call this API directly. Thus, we'll marke this API
+as hidden in IntelliSense. Instead, we'd like to add a compiler feature that
+enables the `fixed` statement to work for any types that offer this method:
```
svick: What is the advantage of not requiring `unsafe`, when it seems the only use case is `fixed`, which does require `unsafe`?

(Also, there's a typo: marke → mark.)

terrajobst: Imagine code like this:

```C#
byte[] bytes = GetBytes();
Span<byte> span = new Span<byte>(bytes);
ref byte value = span.DangerousGetPinnableReference();
value = 42;
```

If `GetBytes()` returns a zero-sized array, the assingment to value
causes memory corruption.

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,247 @@
...
+```C#
+public struct Span<T>
+{
+    public unsafe Span(void* ptr, int length);
+    public Span(T[] array);
+    public Span(T[] array, int start);
+    public Span(T[] array, int start, int length);
+    public static Span<T> Empty { get; }
+    public bool IsEmpty { get; }
+    public ref T this[int index] { get; set; }
```
davidfowl: No setting is required.

terrajobst: Good catch!

2016/11-04-SpanOfT/README.md
svick: My confusion was: why not introduce a method that directly returns a pointer, instead of one that returns `ref T`? But now I understand that such value couldn't be used in `fixed`, since it's already a pointer.

Thanks for the explanation.

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+Status: **Approved** | [Issue](https://github.com/dotnet/corefxlab/issues/951)
+
+We agreed on the following API shape for `Span<T>`:
+
+```C#
+public struct Span<T>
+{
+    public unsafe Span(void* ptr, int length);
+    public Span(T[] array);
```
davidfowl: @jkotas Did we decide to remove the `(ref T, int length)` overload?

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,247 @@
...
+convention and use the term `Dangerous` in the API.
+
+We don't expect user code to call this API directly. Thus, we'll make this API
+as hidden in IntelliSense. Instead, we'd like to add a compiler feature that
+enables the `fixed` statement to work for any types that offer this method:
+
+```C#
+Span<byte> buffer = ...;
+
+fixed (byte* pBuffer = &buffer)
```
davidfowl: We shouldn't need the &. I would assume the compiler would do this for us.

/cc @jaredpar @jkotas

jods4:
What is the difference between `DangerousXXX` and `UnsafeXXX`?
A quick search in .NET sources turns up 310 `UnsafeXXX` and 22 `DangerousXXX`.

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+## `Span<T>` and 64-bit
+
+It was suggested to allow `Span<T>` to represent spans that are larger than
+32-bit.
+
+Before discussion the options it's important to consider that even on 64-bit
+machines, most apps should (and thus will) run as 32-bit because they don't need
+a 64-bit address space and will therefore benefit from smaller working sets as
+the pointers aren't inflated.
```
stephentoub: > even on 64-bit machines, most apps should (and thus will) run as 32-bit

Seems like a very Windows-centric statement.

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,247 @@
+# API Review for `Span<T>`
+
+Status: **Approved** | [Issue](https://github.com/dotnet/corefxlab/issues/951)
+
+We agreed on the following API shape for `Span<T>`:
```
jaredpar: initial shape?

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+Before discussion the options it's important to consider that even on 64-bit
+machines, most apps should (and thus will) run as 32-bit because they don't need
+a 64-bit address space and will therefore benefit from smaller working sets as
+the pointers aren't inflated.
+
+In order to support 64-bit spans we have these options:
+
+1. **Only expose 64-bit APIs**. This would mean having the indexer and `Length`
+   would return `long`.
```
jaredpar: s/return/use

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,247 @@
...
+    public void CopyTo(Span<T> destination);
+    [Obsolete("This operation is not supported.", IsError=True)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public override bool Equals(object obj);
+    [Obsolete("This operation is not supported.", IsError=True)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public override int GetHashCode();
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public ref T int DangerousGetPinnableReference();
+    public static Span<T> DangerousCreate(object obj, IntPtr offset, int length);
```
davidfowl:
```C#
public static Span<T> DangerousCreate(object obj, ref T pointer, int length);
```

```C#
var array = new int[10];

Span<int>.DangerousCreate(array, ref array[2], 8);
```

2016/11-04-SpanOfT/README.md
terrajobst: @jaredpar, what are your thoughts? My asssumption is that you guys want it to look like any other fixed statement.

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,247 @@
...
+{
+    public unsafe Span(void* ptr, int length);
+    public Span(T[] array);
+    public Span(T[] array, int start);
+    public Span(T[] array, int start, int length);
+    public static Span<T> Empty { get; }
+    public bool IsEmpty { get; }
+    public ref T this[int index] { get; }
+    public int Length { get; }
+    public void CopyTo(Span<T> destination);
```
terrajobst: Ah! Good catch. Only in the text 🤘

2016/11-04-SpanOfT/README.md
terrajobst: That's what we said during the review. Not true on Linux? I assume the issue with pointer sizes isn't specific to Windows.

2016/11-04-SpanOfT/README.md
jkotas: The existing equivalent fixed statements for strings and arrays do not have the `&`: `fixed (char * p = myString)` or `fixed (byte * p = myByteArray)`. The generalization should not have it either.

2016/11-04-SpanOfT/README.md
jkotas: Right, this is how the constructors with array arguments are approximately implemented internally. The above is equivalent to much preffered `new Span<int>(array, 2)`.

2016/11-04-SpanOfT/README.md
terrajobst: That's right. Thanks for asking!

2016/11-04-SpanOfT/README.md
terrajobst: Got it. Will leave it off then.

2016/11-04-SpanOfT/README.md
jkotas: In non-Windows world, the simplicity of running everything as 64-bit seems to out-weight the potential performance advantage from using 32-bit in some cases.

There is a way [how to run 32-bit programs on 64-bit linux](http://www.bing.com/search?q=how+to+run+32-bit+programs+on+64-bit+linux), but it is not installed by default (different from Windows) and I have not seen anybody using it for cloud workloads (as always, you can probably somebody who does that).

jkotas: And on Windows, things are moving towards 64-bit as well because of it is more secure (address space randomization, no 32-bit legacy compat) and has smaller footprint if you can switch to it completely. The most used apps - Web Browsers - are 64-bit by default; Windows Nano does not support running 32-bit apps; ... .

terrajobst: Sounds to me like I should simply delete the entire paragraph then.

terrajobst:
> What is the difference between DangerousXXX and UnsafeXXX?

There probably isn't any -- we've prior art for either.

jkotas:
> A quick search in .NET sources turns up 310 UnsafeXXX and 22 DangerousXXX.

Thanks for collecting the data. Sounds like UnsafeXXX is way more common naming pattern and may be  a better choice here.

janvorli:
@jods4, @jkotas I would also prefer the "UnsafeXXX" naming.

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,245 @@
...
+    public bool TryCopyTo(Span<T> destination);
+    [Obsolete("This operation is not supported.", IsError=True)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public override bool Equals(object obj);
+    [Obsolete("This operation is not supported.", IsError=True)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public override int GetHashCode();
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public ref T int DangerousGetPinnableReference();
+    public static Span<T> DangerousCreate(object obj, ref T rawPointer, int length);
```
tmds: What does "obj" do?

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+Since `Span<T>` will have to be special cased by the C# compiler we believe it's
+better to support `foreach` by emitting a regular `for` loop (akin to what the
+compiler does for `string`). We could even generalize this and make it a general
+pattern so that any type that offers an `int`-based `Length` and indexer will
+become enumerable with `foreach`.
+
+Also, it's worth pointing out that there is no point in implementing the
+`IEnumerable<T>` interface as both boxing a instatiating a generic type or
+method with `Span<T>` will be disallowed by the type system.
```
tmds: How is this disallowed by the type system?

terrajobst: The C# compiler will ensure that `Span<T>` cannot be boxed, which includes boxing. By extension, this means disallowing instantiating generic types or methods with `Span<T>`.

2016/11-04-SpanOfT/README.md
davidfowl: There are 2 span implementations, one that can run on any runtime without special support and one that the runtime understands, we'll call it portable span and native span. The native span has 2 fields (pointer and length) and portable span has 3 (object, pointer, length). The native span supports an interior pointer to a managed object and the gc understands how to keep the object alive even though it doesn't have a direct reference to that object. This doesn't work on the portable span, so we need to support passing the object so that both scenarios can work.

jkotas: It is for low-level unsafe tricks that reinterpret memory as `Span<T>`. For example:

```
[StructLayout(LayoutKind.Sequential)]
struct TwoInts { public int X; public int Y; };

class MyClass {
    TwoInts _twoInts;

    ...
        var span = Span<int>.DangerousCreate(this, ref _twoInts.X, 2);
    ...
}
```

davidfowl:
@jkotas with this API how would you make a span from a string?

jkotas:
Something like this would work on CoreCLR (it depends on implementation details that are not part of the ECMA spec though):
```
static ReadOnlySpan<char> ToReadOnlySpan(this string s)
{
    fixed (char * p = s) // There is public method that lets you do this without temporary pin
       return ReadOnlySpan<char>.DangerousCreate(this, ref *p, s.Length);
}
```

davidfowl:
>  creating Span from string is going to be part of the Span APIs.

Yep, that's what I was getting at. We'll need to expand the API for strings and make it so that it can be done with the portable span.

Aside: Wouldn't this work as well?

```C#
static ReadOnlySpan<char> ToReadOnlySpan(this string s)
{
    return ReadOnlySpan<char>.DangerousCreate(this, ref Unsafe.As<StringData>(s).Data, s.Length);
}

private class StringData
{
    public int Length;
    public char Data;
}
```

jkotas:
It would work on CoreCLR today, but it is even more fragile - makes assumptions about the internal implementation details that may not hold in future.

tmds:
Would a Cast operation make sense?
```
public Span<U> Cast(int offset, int length);
```
e.g.
```
var intSpan = byteSpan.Cast<int>(3, 2); // intSpan starts at offset 3 in byteSpan and has Length 2
```

2016/11-04-SpanOfT/README.md
```diff
@@ -0,0 +1,245 @@
...
+```C#
+public struct Span<T>
+{
+    public unsafe Span(void* ptr, int length);
+    public Span(T[] array);
+    public Span(T[] array, int start);
+    public Span(T[] array, int start, int length);
+    public static Span<T> Empty { get; }
+    public bool IsEmpty { get; }
+    public ref T this[int index] { get; }
```
mattwarren: Ah of course, I forgot about the `ref T` bit! You're right, the `{ set; }` isn't needed

jods4: BTW, will it be an error in C# 7 to have a `ref` returning property with a setter? If not, what's the semantic? It seems to me that the same syntax `x[4] = 2` can be interpreted in two different ways (assignment to `ref` or setter call?).

svick: @jods4 It seems it will be an error, I'm currently getting:

> error CS8147: Properties which return by reference cannot have set accessors

PR #35, opened 2017-06-19
Add 2017/06-16-System.Text.Buffers

terrajobst:
@KrzysztofCwalina @ahsonkhan @shiftylogic

### DIFF
--- a/2017/06-16-System.Text.Buffers/README.md
+++ b/2017/06-16-System.Text.Buffers/README.md
@@ -1,0 +1,29 @@
+# API Review for `System.Buffers.Text`
+
+Status: **Needs more work** | [API Reference](System.Text.Buffers.md)
+
+## Feedback
+
+* The assembly name should match the namespace; we have to rename anyway b/c
+  they cause collisions between the CoreFxLab's MyGet feed and NuGet.org
+* We should consider merging the contents of this assembly with the existing
+  assembly
+    - This would allow re-implementing the existing encoding APIs to sit on top
+      of the new APIs
+    - But being out-of-band has some benefit has some benefit, too
+* Maybe we should have a more generic name so that we can put all
+  transformations in that assembly, instead of just APIs related to text
+* `TransformationStatus ` is quite long -- can we make it shorter?
+    - `TransformStatus`?
+* The `ConvertFrom*` and `Compute*` are too long; makes common call sites quite
+  verbose, especially considering the nested classes
+* We've some concerns with the name and shape of the API. Are we taking the good
+  name here? One proposal is to make the namespace more generic, like
+  `System.Buffers.Transformations`
+* `ComputeDecodedLength` -> `GetLength`
+* `ComputeEncodedBytesFromUtf32` -> `GetLengthFromUtf32`
+* `ConvertFromUtf32` -> `FromUtf32`
+* Remove `IBufferFormattable` not needed for now
+* Why do we need `TextFormat`? Shouldn't we be using the existing
+  `NumberFormatInfo`
+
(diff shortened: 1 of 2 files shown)


davkean:
I added the following comments:

What's a default(TextFormat) indicate? Typically we call this "empty", rather than default.
Parsers/Formatters, etc are too general names, and will clash with other (user-defined) types, pick two word compound names.
Nested static class pattern is not really one that we've used before. Usually it's static class with static properties (think Encoding.Unicode, etc). Should these parsers be instance classes similar to encodings?

omariom:
@davkean 
As I understand it this api is expected to be the lowest level one that could be used directly by the end users or by  higher level abstracts. So no  `this` param, no virtual methods etc.

2017/06-16-System.Text.Buffers/System.Text.Buffers.md
```diff
@@ -0,0 +1,178 @@
+```C#
+namespace System.Buffers.Text {
+    public static class Base64 {
+        public static int ComputeDecodedLength(ReadOnlySpan<byte> source);
```
nietras: `LengthOf` perhaps, so becomes `Base64.LengthOf(span)`

2017/06-16-System.Text.Buffers/System.Text.Buffers.md
```diff
@@ -0,0 +1,178 @@
...
+        public static int ComputeEncodedLength(int sourceLength);
```
nietras: does this return length in bytes? or in Base64? is source length in Base64 or? seems a bit too easy to confuse, but I am just seeing this without prior knowledge.

2017/06-16-System.Text.Buffers/README.md
```diff
@@ -0,0 +1,29 @@
...
+  transformations in that assembly, instead of just APIs related to text
+* `TransformationStatus ` is quite long -- can we make it shorter?
+    - `TransformStatus`?
+* The `ConvertFrom*` and `Compute*` are too long; makes common call sites quite
+  verbose, especially considering the nested classes
+* We've some concerns with the name and shape of the API. Are we taking the good
+  name here? One proposal is to make the namespace more generic, like
+  `System.Buffers.Transformations`
+* `ComputeDecodedLength` -> `GetLength`
+* `ComputeEncodedBytesFromUtf32` -> `GetLengthFromUtf32`
```
nietras: agree that `Compute` is superfluous. `GetLengthFromUtf32` does not specify what length is in, aka bytes or in the enclosing type e.g. `Utf16`. Parameter name does but harder to read e.g. `Encoders.Utf16.GetLengthFromUtf32(span, out int length)` is easy to misinterpret. `Get` is also superfluous and rather java like. `Encoders.Utf16.ByteLengthFromUtf32(span, out int byteLength)` 

Even then I am confused what I am asking, I presume what will the byte length of the source span (encoded as Utf32) be if this is encoded as Utf16?

2017/06-16-System.Text.Buffers/README.md
```diff
@@ -0,0 +1,29 @@
...
+* `ConvertFromUtf32` -> `FromUtf32`
```
nietras: the whole `Utf16.FromUtf32` seems weird given the signature, shouldn't it be `Utf16.TryEncodeUtf32` for example, all methods either **encode**, **decode** or get **encoded length of** or **decoded length of** something to/from a given encoding.

2017/06-16-System.Text.Buffers/System.Text.Buffers.md
```diff
@@ -0,0 +1,178 @@
+```C#
+namespace System.Buffers.Text {
+    public static class Base64 {
```
nietras: why is `Base64` not similar to `Utf8` and the other and in namespace and not under `Encoders`?

PR #38, opened 2017-07-17
Add 2017-07-17-Span across CoreFx

terrajobst:
@bartonjs @weshaggard @mellinoe @danmosemsft @stephentoub

### DIFF
--- a/2017/07-17-Span across CoreFx/README.md
+++ b/2017/07-17-Span across CoreFx/README.md
@@ -14,10 +14,15 @@ across CoreFX (with some implementations in CoreCLR/CoreRT).
 
 ## Notes
 
-* Consistency in naming
-    - If we add logical overloads to existing types, we should stick to
-      existing naming conventions on the type
-* BitConverter
+* Principles:
+    - Consistency in naming: if we add logical overloads to existing types, we
+      should stick to existing naming conventions on the type
+    - Keep existing behavior of the types as much as possible
+    - Logically, provide overloads for byte[]
+* `BitConverter`
+    - Should take endianess into account and provide overloads that allows
+      consumers to specify the endianess they want
+    - Should offer a method that allows arbitrary Ts to be read/written
     - `TryCopyBytes`
         + Should be called `TryWriteBytes`
         + Then naming the first argument `value` makes sense. If we stick
--- a/2017/07-17-Span across CoreFx/README.md
+++ b/2017/07-17-Span across CoreFx/README.md
@@ -64,20 +64,24 @@ across CoreFX (with some implementations in CoreCLR/CoreRT).
 * `Guid`
     - `TryCopyTo` should be `TryWriteBytes(Span<bytes> destination)`
 * `String`
-    - The constructor taking the delegate is fine. Ideally, we'd use `Action<TState, Span<char>>` but
-      `Span<T>` cannot be used a generic arguments, so we have to create a custom delegate type.
-      We should probably make this a nested type or put it in a different namespace.
-    - We could make a set of `Func<>` and `Action<>` that already uses `Span<T>` in one of the signatures
-    - Right now, the compiler doesn't allow lambdas and `Span<T>` in the same scope,
-      regardless of whether it's captured or not.
+    - The constructor taking the delegate is fine. Ideally, we'd use
+      `Action<TState, Span<char>>` but `Span<T>` cannot be used a generic
+      arguments, so we have to create a custom delegate type. We should probably
+      make this a nested type or put it in a different namespace.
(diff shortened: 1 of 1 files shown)


2017/07-17-Span across CoreFx/README.md
```diff
@@ -0,0 +1,80 @@
...
+          with `TryCopyBytes` we should probably name it `source`.
+        + `bytes` should be called `destination`
+        + `destination` should be the first argument
+        + `value` should be second
+        + This would also allow us to make them extension methods if wanted
+    - Base-64:
+        + `Try` doesn't work super well as we have two errors: not enough space
+          and invalid data. Invalid data would still throw. We should probably
+          follow CoreFxLab and use an enum. Ideally, we'd use the same, so we
+          should move `TransformationStatus` down:
```
stephentoub: But I think we need a better name.  If nothing else, TransformResult or something like that, though even then I think we may find cases where the operation in question isn't actually a "transform", e.g. just reading data from a file into a a span.

2017/07-17-Span across CoreFx/README.md
```diff
@@ -0,0 +1,80 @@
...
+            public static TransformationStatus ToBase64Chars(ReadOnlySpan<byte> bytes, Span<char> chars, out int bytesConsumed, out int charsWritten, Base64FormattingOptions options = Base64FormattingOptions.None);
+            ```
+* `Random`
+    - The API shape is fine; the question is what the implementation does. We
+      don't want to call the existing array one, because perf tanks. We'll
+      call `Sample` as that's the true primitive and doesn't allocate.
+* Primitive parse
+    - We'll use `ReadOnlySpan<chan>` as a slicable string, always in UTF16
+    - We'll use `TryFormat` as span-based `ToString`
+    - For cases where `ParseExcat` exists, we should offer `Span` based versions
```
jkotas: Typo: ParseExcat

2017/07-17-Span across CoreFx/README.md
```diff
@@ -0,0 +1,80 @@
...
+    - We could make a set of `Func` and `Action` that already uses `Span<T>` in one of the signatures
+    - Right now, the compiler doesn't allow lambdas and `Span<T>` in the same scope,
+      regardless of whether it's captured or not.
+    - Maybe we should make this a separate issue.
+* `Stream`
+    - Sync overloads take `Span<byte>`
+    - Async overloads take `Buffer<byte>`
+    - The implementation on `Stream` will call the existing ones that take `byte[]`,
+      but we'll probably acquire the array from the pool
+    - Out our own derivatives of `Stream` will override the new method to do something smarter
```
jkotas: Typo: Out our own

2017/07-17-Span across CoreFx/README.md
terrajobst: That's fair. The higher order bit for is that we want to use the same type as in CoreFXLab. That's similar to how we use `Buffer<T>` throughout this doc although there is still talk to rename this. Could you file a bug against CoreFXLab to discuss `TransformationStatus`?

stephentoub: > Could you file a bug against CoreFXLab to discuss TransformationStatus?

I can, but what's the split between corefx and corefxlab here?  If we're talking about this for an official API in corefx, doesn't that mean we'd settle on it as part of these discussions?

2017/07-17-Span across CoreFx/README.md
```diff
@@ -0,0 +1,89 @@
...
+            ```C#
+            public static TransformationStatus FromBase64(ReadOnlySpan<char> s, Span<byte> destination, out int charsConsumed, out int bytesWritten, Base46FormattingOptions options = ...);
+            public static TransformationStatus FromBase64(string s, Span<byte> destination, out int charsConsumed, out int bytesWritten, Base46FormattingOptions options = ...);
+            public static string ToBase64String(ReadOnlySpan<byte> bytes, Base64FormattingOptions options = Base64FormattingOptions.None);
+            public static TransformationStatus ToBase64Chars(ReadOnlySpan<byte> bytes, Span<char> chars, out int bytesConsumed, out int charsWritten, Base64FormattingOptions options = Base64FormattingOptions.None);
+            ```
+* `Random`
+    - The API shape is fine; the question is what the implementation does. We
+      don't want to call the existing array one, because perf tanks. We'll
+      call `Sample` as that's the true primitive and doesn't allocate.
```
stephentoub: I looked at this again.  I believe we should actually be calling Next.  The existing NextBytes(byte[]) calls InternalSample, but that's not public; Next() just calls InternalSample and is public and virtual, so it seems like the best alternative to use.

PR #41, opened 2017-08-15
Add 2017/08-15-Intel Intrinsics

terrajobst:


2017/08-15-Intel Intrinsics/README.md
```diff
@@ -0,0 +1,133 @@
...
+parameter value to be a literal.
+
+The idea is to have a way to mark a parameter (e.g. using `const` syntax or via
+an attribute) and encode the requirement as a `modreq` in metadata. We need to
+bring this feature request to the LDT (C# language design team) as it might
+impact overload resolution.
+
+Some addition comments:
+
+* C++ doesn't honor `modreq` (they will still allow the call)
```
jaredpar: C++ honors at least the `modreq` they emit, `IsConst` for example. IIRC they do not honor any others. I could be wrong about the latter part.

2017/08-15-Intel Intrinsics/README.md
```diff
@@ -0,0 +1,133 @@
...
+* F# doesn't honor it either, but they seem to be open to change that
```
jaredpar: Reference: https://github.com/Microsoft/visualfsharp/issues/3105

2017/08-15-Intel Intrinsics/README.md
```diff
@@ -0,0 +1,133 @@
...
+* C# and VB honor them correctly (meaning it doesn't allow binding to methods
+  unless it understands all applied `modreq`s)
```
jaredpar: Subtle difference but C# / VB will bind to such methods, it's just any such binding will result in an error.

benaadams:
Revisiting

> We should probably add overloads that take Span\<T\> or ref T. Let's go with just Span\<T\> for now

With only Span\<T\> it will have to call the pointer methods which will lead to the internals of the method being

```csharp
fixed (byte* pVectors = &span.DangerousGetPinnableReference())
{
    return SSE2.Load(pVectors);
}
```
Whereas going for the `ref` method will mean the Span\<T\> version being
```csharp
return SSE2.Load(ref span.DangerousGetPinnableReference());
```
So it might be better to have the `ref` overload and avoid `fixed`?

/cc @KrzysztofCwalina

2017/08-15-Intel Intrinsics/README.md
```diff
@@ -0,0 +1,133 @@
...
+checks inside of loops.
+
+In the case of AOT, there are basically two options:
+
+1. **No light-up**. This is logically equivalent to run the JIT for a specific
+   environment and persist the output. This means all the capability checks are
+   effectively evaluated at build time. Code will not light up on newer CPU
+   models with more features.
+
+2. *With light-up**. This would mean the build is producing for a specific CPU
```
akoeplinger: missing an `*`

2017/08-15-Intel Intrinsics/README.md
```diff
@@ -0,0 +1,133 @@
...
+   architecture with dynamic checks for possible architecture set. This allows
+   resulting code, for instance, leverage AVX instructions while still running a
+   CPUs that only support SSE2.
+
+This poses interesting questions no how this impacts the granularity of the
```
akoeplinger: typo: **to** how

2017/08-15-Intel Intrinsics/README.md
```diff
@@ -0,0 +1,133 @@
...
+checks because hot loops should have to evaluate capability checks on every
```
akoeplinger: typo: loops **shouldn't** have to

2017/08-15-Intel Intrinsics/README.md
terrajobst: Sorry, I should have qualified that statement as I did for C# and VB.

2017/08-15-Intel Intrinsics/README.md
terrajobst: Fair enough. Will clarify.

terrajobst:
@jkotas 

This is what I thought:

```C#
public static unsafe Vector256<sbyte> Load(sbyte* vector);
public static unsafe Vector256<byte> Load(byte* vector);
public static unsafe Vector256<sbyte> Load(Span<sbyte> vector);
public static unsafe Vector256<byte> Load(Span<byte> vector);
```

terrajobst:
@benaadams 

I assume you're concerned of the overhead of pinning?

benaadams:
> I assume you're concerned about the overhead of pinning?

Yes and then they are handy methods for operating on an array etc and Span can be written using then with the additional length check; rather than with an fixed block.

```csharp
public static unsafe Vector256<sbyte> Load(sbyte* vector);
public static unsafe Vector256<byte> Load(byte* vector);

public static Vector256<sbyte> Load(ref sbyte vector);
public static Vector256<byte> Load(ref byte vector);

public static Vector256<sbyte> Load(Span<sbyte> vector) 
{ 
    if (vector.Length < Vector256<sbyte>.Length) throw;
    return SSE2.Load(ref vector.DangerousGetPinnableReference());
}
public static Vector256<byte> Load(Span<byte> vector) 
{ 
    if (vector.Length < Vector256<sbyte>.Length) throw;
    return SSE2.Load(ref vector.DangerousGetPinnableReference());
}
```
e.g.
```csharp
SSE2.Load(ref array[0]);
```
rather than
```csharp
fixed (byte* ptr = &array[0])
{
    SSE2.Load(ptr);
}
```
Though `ref byte` to be a sequence is a little weird; but its cropping up in places 
e.g. "Avoid manually pinning spans in crypto interop" https://github.com/dotnet/corefx/pull/23259

benaadams:
Fixed probably brothers me with block scoping than raw performance to be fair 😸

ektrah:
IMO `ref byte` for a sequence is very weird. I would only use it internally (as in https://github.com/dotnet/corefx/pull/23259), not in a public API.

terrajobst:
@KrzysztofCwalina, what your thoughts here? I can go either way.

2017/08-15-Intel Intrinsics/README.md
```diff
@@ -10,12 +10,12 @@ We recently approved a [design proposal][Design] to expose hardware intrinsics.
...
 implementation in IL form but is specialized by the code generator (JIT, AOT).
 
-This allows exposing specialized CPU instructions, such as SIMD or Crc32,
+This allows exposing specialized CPU instructions, such as SIMD or CRC32,
 without having to drop to the assembly level. Developers can write regular code,
 think in expressions and local variables, and the code generator can still apply
 the usual optimizations. This approach differs from our existing SIMD
 infrastructure, especially `Vector<T>`, in that these API are hardware specific.
-In other words, code using it is no longer portable across CPU architectures.
+In other words, code using it is no longer portable between CPU architectures.
```
tannergooding: Is it worth indicating that only code using these intrinsics without providing a software fallback is no longer considered portable?

That is:
* code using intrinsics without a corresponding `IsSupported` check is never portable.
* code using intrinsics with corresponding `IsSupported` checks, but no software fallback is partially-portable (it might work on ARM and Intel, but not on future platform X, or on ARM/Intel hardware that doesn't support those intrinsics).
* code using intrinsics with a software fallback (that does not use intrinsics) should be always portable (or at least as portable as the runtime is)

terrajobst: Good point reworded to:

> This approach differs from our existing SIMD infrastructure, especially
`Vector<T>`, in that these API are hardware specific. In other words, code using
it is no longer portable between CPU architectures. Furthermore, we do not
provide a software fallback. Instead, the developer is expected to guard calls
using a provided capability API. Failing to do so will cause
`PlatformNotSupportedException` at runtime.
>
> In order to remain portable, the developer must provide a software fallback.
While this sounds like we're putting the burden on the developer, we learned
that we simply can't provide an efficient enough software fallback at the
granularity level of intrinsics. Vectorized and non-vectorized approaches are
fundamentally different and running a software-implemented vectorized algorithm
is often several factors slower than a native non-vectorized version.

terrajobst:
Merging, because we've reached the point where we discucss the feature, rather than the meeting notes.

nietras:
I also don't like the new `Span` `Load/Store` we need the `ref T` versions instead and this would support both native pointers and `ref`s from a span naturally and without pinning or hoops around if one has to create a `Span` just to `Load`.

terrajobst:
That's a good point: if we only had pointer based versions and `ref T`-based version, what would a `Span<T>` based version buy us?

benaadams:
> pointer-based version and a ref T-based version

Covers all scenarios. Without `ref T` is only pointer based so needs pinning; also then `Span<T>` and be done in terms of `ref T` without pinning.

For `Span<T>` overloads was conclusion that they wanted to be done in "user code" rather than "runtime code" so they could inline and the bounds checks could be coalesced/eliminated without versioning issues?

nietras:
> Covers all scenarios. Without ref T is only pointer based so needs pinning

Precisely, `Span<T>` could be done based on `ref T` overloads in different scope, and even as extension methods if that would make sense. `Span` overloads benefit from bounds check as you say, but at the cost of efficiency and necessitates creating spans when not really used. Ideally, the intrinsics API would have pointer and `ref T` overloads as I see it. But even pointer overloads could be seen as superfluous since they can easily be done via `ref T` as:
```csharp
float* ptr = nativeArray;
SSE2.Load(ref *ptr);
```
The drawback here would be that this adds a `ref` to the stack that the GC needs to check against segments, unless the JIT can recognize this as being unnecessary.

Additionally, having `Span` overloads creates an unnecessary coupling between intrinsics and `System.Memory`/`Span<T>` which may not be a big concern but any unnecessary coupling is bad in my view.

To sum up, please no `Span<T>` overloads but pointer and `ref T` instead, thank you 😄

benaadams:
Also `LoadAligned(` can only have a pointer version, no ref

As you can only know if its aligned (currently) with a fixed pointer (ref is GC location dependent)

nietras:
> you can only know if its aligned (currently) with a fixed pointer (ref is GC location dependent)

@benaadams  I'm not sure that is true. A `ref` can also reference unmanaged memory and you can in that case know it won't move. Yes, I would in fact use `ref`s for unmanaged memory to avoid having to duplicate code in some cases. Or it can `ref` a stack location, which cannot move either.

benaadams:
> Or it can ref a stack location, which cannot move either.

Why would you be LoadAligned a stack location? Either its already the Vector type; or its stackalloc'd when you have a pointer.

LoadAligned is more specialist function and you shouldn't be not duplicating its use with items you know the location of and items you don't know the location of. As you can only LoadAligned items you know the location of.

nietras:
> Why would you be LoadAligned a stack location? Either its already the
Vector type; or its stackalloc'd when you have a pointer.

Good point ☺️ Think that crushes that argument 😁 Only thing would be in
cases where unsafe code could not be used... but why indeed then...

And since managed arrays only have known alignment up to 8 bytes that
wouldn't help us for 16/32/etc. alignments.

Because you can compute the alignment of a `ref` but you are probably right
it wouldn't be particularly useful it seems.

tannergooding:
C# is adding `stackalloc` support for `Span<T>` (https://github.com/dotnet/roslyn/pull/21006). So, in some future language version, you'll be able to do `Span<T> x = stackalloc` rather than just `T* x = stackalloc`

benaadams:
Still need to look at the pointer to determine alignment; `stackalloc` doesn't offer alignment options.

nietras:
No alignment of a `ref` can also be determined. It is not straightforward
since we can't easily make a zero/null `ref`, but it is possible.
Unsafe.ByteOffset, if I remember name correctly can be used.

I also thought of the Span stackalloc so perhaps in the future a need will
surface and perhaps we should add it then rather than now, since it does
indeed seem a little esoteric although not invalid.

fiigii:
>  Let's go with just `Span<T>` for now  

Shall we change to `ref T`?

benaadams:
> ´Unsafe.ByteOffset´, if I remember name correctly can be used.

It isn't valid to use that for an **absolute** memory location calculation across two statements as its location may have changed. Much like using Unsafe.AsPointer on a non-fixed location and then using the pointer as an absolute location as that location may change between measurement and use.

> Shall we change to `ref T`?

I think `*` and `ref T` cover the maximum possible use cases. On top (outside this change) its then very easy to add `Span<T>`  overloads via `ref T` without fixing.

However `*` and `Span<T>` overloads don't allow you to use unchecked `ref T` overloads; you need instead to fix and use `*` or create very many `Span<T>` and have each call checked (and the `Span<T>` will need to fix and use `*` behind the scenes).

/cc @terrajobst @KrzysztofCwalina

PR #42, opened 2017-09-01
Add API review session of GitHub issues on 2017/8/29

karelz:


terrajobst:
The only you may want to consider is adding a status to the individual items (Approved, Approved with Feedback, Rejected)

PR #58, opened 2018-01-10
Add 2018/System.IO.Pipelines

terrajobst:
These are notes from today's meeting.

@pakrym @halter73 @davidfowl @krwq @dotnet/fxdc


2018/System.IO.Pipelines/README.md
```diff
@@ -0,0 +1,53 @@
...
+  naming collisions.
+
+### `ReadOnlyBuffer`
+
+* Should be generic
+* `Move` should be called `Seek` or `GetPosition`. It seems to cause some
+  confusion as it implies mutation.
+* Lots of `Position` parameters are still called cursor
+* We should add `Slice` overloads that take `int` in addition to `long`
+* `ReadOnlyBuffer` shouldn't take `OwnedMemory<T>` (will be easier after
```
davidfowl: `ReadOnlyBuffer` should take `ReadOnlyMemory<T>` instead of `OwnedMemory<T>` and array overloads

2018/System.IO.Pipelines/README.md
```diff
@@ -0,0 +1,53 @@
+# System.IO.Pipelines Review: `ReadOnlyBuffer` and `IBufferList`
+
+Status: **Needs more work** | 
+[API](System.IO.Pipelines.md) |
+[Presentation](System.IO.Pipelines.pptx) |
+[Video](https://www.youtube.com/watch?v=tLnQsNTfl9Q)
+
+## Intro
+
+* There are are multiple categories of types (see columns in slide #2)
```
ahsonkhan: `are are`

2018/System.IO.Pipelines/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+### `Position`
+
+* `Position` (used to be `ReadCursor`) is a very generic name, maybe we should
+  consider prefixing it with something, e.g. `BufferPosition`, in order to avoid
+  naming collisions.
+
+### `ReadOnlyBuffer`
+
+* Should be generic
```
ahsonkhan: Since it is now generic, APIs that do not need the internals of ReadOnlyBuffer (like PositionOf) should be extension methods. Can we use the exisiting MemoryExtensions class?
https://github.com/dotnet/corefx/blob/master/src/System.Memory/ref/System.Memory.cs#L89

ahsonkhan: Also, are we adding TryCopyTo?

2018/System.IO.Pipelines/README.md
```diff
@@ -0,0 +1,53 @@
+# System.IO.Pipelines Review: `ReadOnlyBuffer` and `IBufferList`
+
+Status: **Needs more work** | 
+[API](System.IO.Pipelines.md) |
```
ahsonkhan: The APIs listed here do not show the relevant APIs that were discussed like ReadOnlyBuffer/Position/IBufferList

2018/System.IO.Pipelines/README.md
```diff
@@ -0,0 +1,53 @@
...
+### `ReadOnlyBuffer`
+
+* Should be generic
+* `Move` should be called `Seek` or `GetPosition`. It seems to cause some
+  confusion as it implies mutation.
+* Lots of `Position` parameters are still called cursor
+* We should add `Slice` overloads that take `int` in addition to `long`
+* `ReadOnlyBuffer` shouldn't take `OwnedMemory<T>` (will be easier after
+  `ReadOnlyBuffer` is moved to `System.Memory.dll`).
+
```
ahsonkhan: Missing feedback: Rename IsSingleSpan property

2018/System.IO.Pipelines/README.md
terrajobst: Good point. I only exported `System.IO.Piplines`. Fixed.

2018/System.IO.Pipelines/README.md
terrajobst: Fixed

2018/System.IO.Pipelines/README.md
terrajobst: Fixed

2018/System.IO.Pipelines/README.md
jkotas: > Can we use the exisiting MemoryExtensions class?

MemoryExtensions should be for the lowest level stuff that is moving to CoreLib. Do you expect the pipelines to live in CoreLib?

2018/System.IO.Pipelines/README.md
terrajobst: > Missing feedback: Rename IsSingleSpan property

Fixed

> Also remove the constructor that takes Position

Which one?

```csharp
        public ReadOnlyBuffer(OwnedMemory<byte> data, int offset, int length);
        public ReadOnlyBuffer(byte[] data);
        public ReadOnlyBuffer(byte[] data, int offset, int length);
        public ReadOnlyBuffer(IEnumerable<Memory<byte>> buffers);
        public ReadOnlyBuffer(IBufferList startSegment, int offset, IBufferList endSegment, int endIndex);
```

2018/System.IO.Pipelines/README.md
```diff
@@ -0,0 +1,53 @@
...
+### `IBufferList`
+
+* Should be generic
+* `IBufferList` is really a `IMemoryList`, so we should either change the
+  property name or the type name (changing the type name seems more logical).
+* `VirtualIndex` seems to be a name folks don't understand. Maybe it should be
+  `GlobalIndex` or maybe even just `Index` or `Offset`.
```
benaadams: Not `GlobalIndex` as that implies whole program

2018/System.IO.Pipelines/README.md
terrajobst: Not pipelines, but we're considering `ReadOnlyBuffer` to `System.Memory` and thus to corlib.

2018/System.IO.Pipelines/README.md
terrajobst: Agreed; I guess `Index` it is.

2018/System.IO.Pipelines/README.md
```diff
@@ -0,0 +1,56 @@
...
+* Should `IBufferList` be read-only? Downside is that we couldn't reuse it for
+  a (hypothetical) type `ReadWriteBuffer` without adding a mutable version of
+  `IBufferList` too.
```
benaadams: Both is better, or it can only be read-only as conversion only goes one way.
e.g.
`IMemoryList<T>` taking `Memory<T>` and `IReadOnlyMemoryList<T>` representing `ReadOnlyMemory<T>` seems to make sense

terrajobst: True. We briefly covered this during the review and I find myself agreeing with @KrzysztofCwalina in that I don't see the value of `IReadOnlyMemoryList<T>` as this type is only used inside the implementation of `ReadOnlyBuffer` and doesn't leak to consumers. To me, it's a building block for higher level types similar to how `List<T>` is the building block for `Collection<T>` and `ReadOnlyCollection<T>`. It seems providing read-only alternatives for those types is going over the top with type safety.

benaadams: Yeah; can't define a terminating, or varying sized stack only lists using it as a generic constraint, so its little value

2018/System.IO.Pipelines/README.md
weshaggard: I thought we are talking about somehow indicating it is a cumulative index which includes the length.

2018/System.IO.Pipelines/README.md
```diff
@@ -0,0 +1,56 @@
...
+  and a single contiguous buffer can logically contain many spans).
+
+### `IBufferList`
+
+* Should be generic
+* `IBufferList` is really a `IMemoryList`, so we should either change the
+  property name or the type name (changing the type name seems more logical).
+* `VirtualIndex` seems to be a name folks don't understand. Maybe it should be
+  `GlobalIndex` or maybe even just `Index` or `Offset`.
+* Should `IBufferList` be read-only? Downside is that we couldn't reuse it for
```
weshaggard: What about the potential name change to IMemoryList?

weshaggard: nevermind I found that below

2018/System.IO.Pipelines/README.md
KrzysztofCwalina: Maybe "RunningIndex", I think this is what it was internally called in pipelines at some point.

2018/System.IO.Pipelines/README.md
KrzysztofCwalina: I think ```ReadOnlyBuffer``` should go to System.Memory, but not to corlib.

2018/System.IO.Pipelines/README.md
pakrym: `RunningLenght`

2018/System.IO.Pipelines/System.IO.Pipelines.md
```diff
@@ -0,0 +1,249 @@
+# System.IO.Pipelines.dll
+
```
KrzysztofCwalina: Why all these APIs we did not review are here? Is it clear that they were not reviewed?

terrajobst: Fair point. I didn't bother removing the ones we didn't. Will strip them.

2018/System.IO.Pipelines/README.md
terrajobst: @weshaggard 

> I thought we are talking about somehow indicating it is a cumulative index which includes the length.

I'm not sure that's relevant. The index is the start where the current node is in the virtual buffer represented but the entire list, in other words the first node will have set this to 0 and the next one to first.Length. I'm almost thinking it's easier to just call it Index as it's really the index of this instance in the virtual buffer.

terrajobst: We may want to consider renaming `IBufferList` to `IBufferNode`. The list if the entirety (and doesn't need a type) while the concept that developers need to understand is really the node. It feels a bit odd not using it as the type.

karelz: What about `OuterIndex`? (not a strong opinion, just throwing in an option)

weshaggard: All I can say is that having multiple things called Index that mean slightly different things is confusing. Reading this interface API in isolation I wouldn't be able to figure out what this index is for without reading the documentation.

benaadams: `ListIndex`

weshaggard: That still is confusing as I would expect it to be the index in the IBufferList so 0,  1, 2, etc not necessarily the index into the aggregate buffer.

2018/System.IO.Pipelines/System.IO.Pipelines.md
```diff
@@ -0,0 +1,68 @@
...
+    public struct ReadOnlyBuffer : ISequence<ReadOnlyMemory<byte>> {
+        public ReadOnlyBuffer(OwnedMemory<byte> data, int offset, int length);
+        public ReadOnlyBuffer(byte[] data);
+        public ReadOnlyBuffer(byte[] data, int offset, int length);
+        public ReadOnlyBuffer(IEnumerable<Memory<byte>> buffers);
+        public ReadOnlyBuffer(IBufferList startSegment, int offset, IBufferList endSegment, int endIndex);
+        public Position End { get; }
+        public ReadOnlyMemory<byte> First { get; }
+        public bool IsEmpty { get; }
+        public bool IsSingleSpan { get; }
```
KrzysztofCwalina: This is now called IsSingleSegment

2018/System.IO.Pipelines/README.md
terrajobst: Agreed. Well, if nobody has a better suggestion, I'm leaning towards `Index` as it (seems) to raise the fewest eyebrows.

Any thoughts on the name of the interface?

2018/System.IO.Pipelines/System.IO.Pipelines.md
```diff
@@ -0,0 +1,68 @@
...
+    public struct Position : IEquatable<Position> {
+        public Position(object segment);
+        public Position(object segment, int index);
+        public int Index { get; }
+        public object Segment { get; }
+        public bool Equals(Position position);
+        public override bool Equals(object obj);
+        public ValueTuple<T, int> Get<T>();
+        public override int GetHashCode();
+        public T GetSegment<T>();
```
KrzysztofCwalina: This is now gone.

2018/System.IO.Pipelines/System.IO.Pipelines.md
```diff
@@ -0,0 +1,68 @@
...
+        public static Position operator +(Position value, int index);
```
KrzysztofCwalina: This is now gone.

2018/System.IO.Pipelines/System.IO.Pipelines.md
terrajobst: I generally leave the APIs as they were brought and leave any changes to the (written) feedback. Otherwise it always ends up in a half-half situation as most changes we agreed upon cannot be done by me but require thought on the API owner's side.

2018/System.IO.Pipelines/System.IO.Pipelines.md
terrajobst: See earlier [comment](https://github.com/dotnet/apireviews/pull/58#discussion_r160801750)

2018/System.IO.Pipelines/System.IO.Pipelines.md
terrajobst: See earlier [comment](https://github.com/dotnet/apireviews/pull/58#discussion_r160801750)

2018/System.IO.Pipelines/System.IO.Pipelines.md
KrzysztofCwalina: kk

2018/System.IO.Pipelines/System.IO.Pipelines.md
```diff
@@ -0,0 +1,68 @@
...
+        public ReadOnlyBuffer(byte[] data, int offset, int length);
+        public ReadOnlyBuffer(IEnumerable<Memory<byte>> buffers);
+        public ReadOnlyBuffer(IBufferList startSegment, int offset, IBufferList endSegment, int endIndex);
+        public Position End { get; }
+        public ReadOnlyMemory<byte> First { get; }
+        public bool IsEmpty { get; }
+        public bool IsSingleSpan { get; }
+        public long Length { get; }
+        public Position Start { get; }
+        public void CopyTo(Span<byte> destination);
```
JamesNK: What about start and count args? Or is the idea that you would do a Slice to create a new ReadOnlyBuffer and then a CopyTo?

davidfowl: Yes, that's the general pattern when you accept a Span\<T\>.

JamesNK: The scenario I was thinking of is what if you want to copy a subset of the `ReadOnlyBuffer` to the destination.

davidfowl: Yes, you'd also need to slice the `ReadOnlyBuffer` itself to enable that scenario.

2018/System.IO.Pipelines/System.IO.Pipelines.md
terrajobst: It's [here](https://github.com/dotnet/apireviews/pull/58/files#diff-acd2bd8765fe235625ab96076b836efeR38)

---
In reply to: [160808056](https://github.com/dotnet/apireviews/pull/58#discussion_r160808056) [](ancestors = 160808056)

2018/System.IO.Pipelines/README.md
terrajobst: If we can reuse an existing type, I'm not opposed (but I don't think it's super advantageous either). However, I really don't think we should be adding another purely to hold the extension methods.

---
In reply to: [160568829](https://github.com/dotnet/apireviews/pull/58#discussion_r160568829) [](ancestors = 160568829)

terrajobst: Sounds good.

---
In reply to: [160743847](https://github.com/dotnet/apireviews/pull/58#discussion_r160743847) [](ancestors = 160743847)

terrajobst: @ahkha@microsoft.com 

> Also, are we adding TryCopyTo?

Good question.

---
In reply to: [160568849](https://github.com/dotnet/apireviews/pull/58#discussion_r160568849) [](ancestors = 160568849)

pakrym: Maybe in next iteration. It would be extension method anyway.

2018/System.IO.Pipelines/README.md
ahsonkhan: I was referring to the one that took two Positions, which has already been removed.

2018/System.IO.Pipelines/README.md
pakrym: `IMemoryListNode<T>`? People seem to be confused that it's a pointer to the node, not the entire list.

2018/System.IO.Pipelines/README.md
terrajobst: Excellent.

---
In reply to: [161108505](https://github.com/dotnet/apireviews/pull/58#discussion_r161108505) [](ancestors = 161108505)

2018/System.IO.Pipelines/README.md
terrajobst: I like it.

halter73: I think the reason people are opposed to just `Index` is because it's natural to assume that it indexes directly into the node's Memory which it does not.

I haven't heard any objections to `RunningLength`.

2018/System.IO.Pipelines/README.md
```diff
@@ -0,0 +1,56 @@
...
+    - `IBufferList`
+* `Span<T>`/`Memory<T>` friends are for contiguous memory; pipelines add
+  primitives that handles discontinuous memory. The analogy that
+  [@Krzysztof](https://github.com/KrzysztofCwalina) provided was:
+    - `ReadOnlyBuffer` ~ `Memory<byte>` (but discontinuous)
+    - `Pipe` ~ `System.Threading.Channel<byte>`
+* Writing is generally easier and thus has fewer types; thus full API-by-API
+  symmetry between reading and writing might be hard.
+* .NET has multiple layers of buffers, many of which are implementation details
+  of a particular class. Pipelines makes makes buffers a first-class concern to
```
nietras: nit: "makes makes" => make

2018/System.IO.Pipelines/README.md
terrajobst: Feels a bit off to me, but given nobody objects I'm fine with that too.

---
In reply to: [161110556](https://github.com/dotnet/apireviews/pull/58#discussion_r161110556) [](ancestors = 161110556)

PR #59, opened 2018-01-13
Add 2018/System.IO.Pipelines_2

terrajobst:
This isn't complete.

@halter73 @pakrym: please help me with explanation on the [three modes](https://github.com/dotnet/apireviews/compare/master...terrajobst:pipelines2?expand=1#diff-9a2486d8ede3f7087be0fe230bf00decR34). I'm also unclear which decision we made.


2018/System.IO.Pipelines_2/README.md
```diff
@@ -0,0 +1,91 @@
...
+}
+```
+
+* Three modes for sending data to the reader
+    1. `FlushAsync`. Will send the data to the reader and the reader will wake
+       up from the `ReadAsync` call
+    2. Commit.
+    3. Partial flush. Currently not supported. Do we need this? Ben and Tim
+       seemed to believe it's useful. Pavel thinks it can added a later. The
+       workaround would be a side channel (variable + lock).
```
benaadams: > * Three modes for sending data to the reader

Aren't really modes; as each step is a cumulative stage on each other
1. Commit - Transfers ownership of data (e.g. its committed, can't be changed by writer)
   * Does not notify reader
   * Reader can opportunistically pick up, if its operating in parallel (pipe as thread comms)
   * Also supports non-blocking readers, as no data would indicate a "read would block state" @tmds 
2. NotfiyAsync -> (Commit if not done), inform reader of data, can receive back pressure
3. FlushAsync -> (Commit+Notify) additionally says to flush internal buffer to output

2 & 3 could be merged as essentially the difference is a flag

`FlushAsync(bool moreData = false)`

Where:
* `NotifyAsync` == `FlushAsync(moreData: true)`
* `FlushAsync` == `FlushAsync(moreData: false)`

`FlushAsync(moreData: true)` wakes up the next stage, but allows it to determine its own framing. Whether that's waiting till a TCP packet is full before sending, or a 16kB TLS frame is full before closing and starting another (examples https://github.com/dotnet/corefxlab/issues/2029#issuecomment-355790366)

Whereas `FlushAsync(moreData: false)` says no point in waiting is no more data immediately coming so send everything you can. Operates slightly like Nagel, except Nagel is at the TCP layer and doesn't know anything about the data state in the upper layers so has to use a timeout; hence have to switch it off for low latency.

Example writing a file from ASP.NET current:
* Loop till end of file
  1. Read N kB from file 
  2. Write All to Pipe 
  3. Flush All 
  4. Read All from Pipe 
  5. Open Compression frame, Compress All, Close frame, and Write All to Pipe 
  6. Flush All
  7. Read All
  8. Open TLS frame, Encrypt All, Close TLS frame, and Write All to Pipe
  9. Flush All
  10. Read All -> *syscall*
  11. Start TCP packet, Write All to packet, Send packet

Every layer and frame and number of *syscall*s is determined by the choice of buffer size in **i.**; which is bad as it doesn't work with the natural sizes of each pipe step, opening an closing frames/packets adds to data bloat; also doubly bad with Meltdown and Spectre as *syscall*s now can have a +20% extra cost.

Instead, if they use `FlushAsync(moreData: true)`  frames can be left open; packets only sent when they are full etc; then the `FlushAsync(moreData: false)` happens when there is no more immediate input data at the start of the pipe and it can cause the entire pipeline to flush and close any partial frames/packets and send them.

Why not just notify and flush, with no commit? `Commit` is important for parallel processing; as notify and flush are forced synchronization points; also `Commit` is used in deeper sync stacks; whereas Notify/Flush are used in the async outer stacks.

Precedent `StreamWriter`->`void Flush(bool flushStream, bool flushEncoder)`

`flushEncoder` says write internal buffer to the stream and clear it.
`flushStream` says flush the underlying data you've written.

Getting started samples would only need to show `FlushAsync` which would use default param.

Help any?

benaadams: *Ninja edit to add:

Currently each pipe stage in the pipeline **must** itself do a full flush as they don't know if there is going to be more data of if they will block forever waiting for it.

benaadams: For ASP.NET its only the server's HTTP parser needs to force a full flush, when the current request has finished; and there is not enough data on the input socket to start a new request (e.g. end of headers for new request)

benaadams: *Ninja edit 

For ASP.NET its only the server's HTTP parser that needs to force a full flush, when the current request has finished; and there is not enough data on the input socket to start a new request (e.g. end of headers for new request)

*Websockets, Server Sent Events are exceptions which would need to force full flushes from the application layer - either explicit Flush (SSE) calls or on message complete for WS. 

E.g. At application layer 

`WriteAsync` should call `FlushAsync(moreData: true)` 
`FlushAsync` should call  `FlushAsync(moreData: false)` 

Currently `WriteAsync` would be calling `FlushAsync(moreData: false)`  as there isn't a differentiation between the two.

/cc @halter73 @pakrym @davidfowl

benaadams: Putting it another way:

Currently `WriteAsync` would be calling `FlushAsync(moreData: false)`  as there isn't a differentiation between a `WriteAsync` and a `WriteAsync`+`FlushAsync` as there is no method to call on the Pipe that differentiates them.

2018/System.IO.Pipelines_2/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Write Loop
+
+```csharp
+while (true)
+{
+    try
+    {
+        var memory = pipe.Writer.GetMemory();
+        // ... wrire to memory
```
benaadams: write

2018/System.IO.Pipelines_2/README.md
```diff
@@ -0,0 +1,91 @@
...
+* `Advance(consumed)` shouldn't dead-lock or cause inifinite loops
+    - Calls `Advance(consumed, consumed)` unless `consume` is `0`, in which case
+      it will call `Advance(0, reader.Buffer.End)`
+* `IPipeConnection`
+    - It seems folks would like to have *duplex* in the name, but that would
+      make the name long (`IDuplexPipeConnection`)
+    - Sounds like we've settled on `IPipeConnection`
+* `PipeOptions`
+    - Should expose a static readonly property `Default` that uses the default memory pool
+    - `maximumSizeLow` -> `backPressureEnds`
```
davidfowl: These names aren't great. If we're looking at other framework comparisons, netty uses high watermark and low watermark:

```
WRITE_BUFFER_HIGH_WATER_MARK
WRITE_BUFFER_LOW_WATER_MARK
```

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
```diff
@@ -0,0 +1,69 @@
...
+    public long MaximumSizeHigh { get; }
+    public long MaximumSizeLow { get; }
+    public int MinimumSegmentSize { get; }
+    public MemoryPool<byte> Pool { get; }
+    public Scheduler ReaderScheduler { get; }
+    public Scheduler WriterScheduler { get; }
+} 
+
+public static class PipelineExtensions {
+    public static Task WriteAsync(this IPipeWriter output, byte[] source);
```
davidfowl: We would remove this one right? We don't need both byte[] and ReadOnlyMemory\<byte\>

2018/System.IO.Pipelines_2/README.md
benaadams: `PauseWriterThreshold`
`ResumeWriterThreshold`

benaadams: Or lots of words version:

`PauseWriterBacklogThreshold`
`ResumeWriterBacklogThreshold`

2018/System.IO.Pipelines_2/README.md
omariom: @benaadams 

Thanks to this dicussion I better undestand the machinery.

> NotfiyAsync -> (Commit if not done), inform reader of data, can receive back pressure

So `NotfiyAsync` is "notify the reader" and `FlashAsync` is like `NotfiyAsync` folowed by a soft order to the reader to 'FlashAsync' to the next stage?

benaadams: Yes, but `FlushAsync` is a hard order to push all of its data to the next stage (it may be many minutes till more data; e.g. next send on a WebSocket) and to `FlushAsync` that stage, and that stage to `FlushAsync` etc

benaadams: `System.Web.HttpResponse` has a similar internal differentiation: 
`private void FinalFlushAtTheEndOfRequestProcessing(bool needPipelineCompletion)`
`private void Flush(bool finalFlush, bool async = false)`

tmds: Does this require that the reader can differentiate between being woken up by NotifyAsync vs FlushAsync? In the first case it may decide to wait a little for more data?

What if `WriteAsync` wouldn't call `FlushAsync` each time, but only when the pipe is too full?

benaadams: > Does this require that the reader can differentiate between being woken up by NotifyAsync vs FlushAsync? In the first case it may decide to read for more data?

Yes; or it might Write some (e.g. 2 packets full); then Read waiting for more (e.g. hold onto 0.5 packet)

> What if WriteAsync wouldn't call FlushAsync each time, but only when the pipe is too full?

e.g. it only calls `Commit` mostly? Its buffering in the wrong place its the next stage that knows about the remaining levels in the next stages buffers; the prior writer has no knowledge of what the correct Flush points are for the next stage. Also a forced flush will likely Flush every stage.

(Though in practice it would likely do FlushAsync->Process->Write->Recheck for more data->If none Force flush on next stage)

benaadams: Being in backpressure should probably also convert a Notify to a Flush; else the pipeline would lock (bad if you had your numbers wrong)

tmds: > the prior writer has no knowledge of what the correct Flush points are for the next stage.

What is the writer's intent when he calls 'WriteAsync'?
I think it's to append some data, and get stopped if the Pipe is too full.

> Its buffering in the wrong place its the next stage that knows about the remaining levels in the next stages buffers

If WriteAsync doesn't notify the reader each time, it is buffering in the Pipe. Is that bad?

benaadams: > If WriteAsync doesn't notify the reader each time,

The issue is you only have `Commit` and `FlushAsync` there is currently no equivalent of `WriteAsync` (e.g. Commit + Notify) [so its done using a full flush via `FlushAsync`](https://github.com/dotnet/corefxlab/blob/b5a998d2758b8ee968dad324f104c9310057eda3/src/System.IO.Pipelines/System/IO/Pipelines/PipeExtensions.cs#L18-L36) - which is a different behavior

2018/System.IO.Pipelines_2/README.md
```diff
@@ -0,0 +1,91 @@
...
+    - `maximumSizeHigh` -> `backPressureStarts`
+* `Pipe`
+    - Offer an constructor with notarguments that passes in
+      `PipeOptions.Default`
+* `IOutput`
+    - In order to settle on `IOutput` as the key constraint for writers, we need
+      a way get a `Span<T>` as this is often more efficient.
```
mgravell: Possibly related: https://github.com/dotnet/corefxlab/issues/2040

2018/System.IO.Pipelines_2/README.md
tmds: > I'm suggesting there should be an intermediary between Commit and FlushAsync which behaves like WriteAsync

I'm asking if this can't be internal to WriteAsync, i.e. no additional public API for reader and writer.

benaadams: Could be something like remove `FlushAsync` and add:
```csharp
public interface IPipeWriter : IOutput {
    ValueAwaiter<WriteResult> WriteAsync(bool flush = true, CancellationToken cancellationToken = default);
}
```
Keep
```csharp
public static class PipelineExtensions {
    public static Task WriteAsync(this IPipeWriter output, ReadOnlyMemory<byte> source);
}
```
Change `ReadResult` to:
```csharp
[Flags]
public enum ReadStatus : byte
{
    IsCancelled,
    IsCompleted,
    IsFlush
}

public struct ReadResult {
    public ReadResult(ReadOnlyBuffer buffer, bool isCancelled, bool isCompleted);
    public ReadOnlyBuffer Buffer { get; }
    public ReadStatus Status { get; }
}
```

halter73: I'm not completely sold on NotifyAsync yet. The `ReadStatus.IsFlush` seems unintuitive to me.

benaadams: Not sure about naming `NotifyAsync` but a combined method that takes a flag so you could do
```csharp
public static class PipelineExtensions
{
    private static readonly Task _completedTask = Task.FromResult(0);

    public static Task WriteAsync(this IPipeWriter output, ReadOnlyMemory<byte> source)
    {
        var writeBuffer = output.Alloc();
        writeBuffer.Write(source.Span);

        var awaitable = writeBuffer.NotifyAsync(flush: false);
        if (awaitable.IsCompleted)
        {
            awaitable.GetResult();
            return _completedTask;
        }

        return FlushAsyncAwaited(awaitable);
    }
     
    public static Task FlushAsync(this IPipeWriter output)
    {
        var awaitable = writeBuffer.NotifyAsync(flush: true);
        if (awaitable.IsCompleted)
        {
            awaitable.GetResult();
            return _completedTask;
        }

        return NotifyAsyncAwaited(awaitable);
    }

    private static async Task NotifyAsyncAwaited(ValueAwaiter<NotifyResult> awaitable)
    {
        await awaitable;
    }
}
```

2018/System.IO.Pipelines_2/README.md
terrajobst: @davidfowl @KrzysztofCwalina @halter73 @pakrym so what's the conclusion regarding these names?

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
terrajobst: I'm OK with that. @KrzysztofCwalina?

2018/System.IO.Pipelines_2/README.md
KrzysztofCwalina: I like PauseWriterThreshold and ResumeWriterThreshold (and the docs clarifying that the unit is bytes)

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
KrzysztofCwalina: Yeah, there is an implicit cast from byte[] to ROM

2018/System.IO.Pipelines_2/README.md
halter73: 👍 for @benaadams orignal suggestion.

davidfowl: Yep

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
```diff
@@ -0,0 +1,69 @@
...
+    public static Task WriteAsync(this IPipeWriter output, ReadOnlyMemory<byte> source);
+    }
+
+public interface IOutput {
```
KrzysztofCwalina: Since we decided to change ReadOnlyBuffer to be over T, should this interface also be generic?, i.e. IOutput<T>? And while we are at it, it would be nice if we called it IBufferWriter<T>

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
```diff
@@ -0,0 +1,69 @@
+# System.IO.Pipelines
+
+```csharp
+public interface IPipe {
+    IPipeReader Reader { get; }
+    IPipeWriter Writer { get; }
+}
+public interface IPipeReader {
+    ValueAwaiter<ReadResult> ReadAsync(CancellationToken cancellationToken=default(CancellationToken));
```
jaredpar: Can simplify to `CancellationToken cancellationToken = default` now.

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
```diff
@@ -0,0 +1,69 @@
...
+
+    void Advance(Position consumed);
+    void Advance(Position consumed, Position examined);
+
+    void CancelPendingRead();
+    void Complete(Exception exception=null);
+    void OnWriterCompleted(Action<Exception, object> callback, object state);
+}
+public interface IPipeWriter : IOutput {
+    ValueAwaiter<FlushResult> FlushAsync(CancellationToken cancellationToken=default(CancellationToken));
```
jaredpar: Can simplify to `CancellationToken cancellationToken = default` now.

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
```diff
@@ -0,0 +1,69 @@
...
+    void Commit();
+
+    void CancelPendingFlush();
+    void Complete(Exception exception=null);
+    void OnReaderCompleted(Action<Exception, object> callback, object state);
+}
+
```
jaredpar: Extra new line between types. Don't care which is done but all other types have no blank line between.

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
```diff
@@ -0,0 +1,69 @@
...
+public interface IPipeConnection : IDisposable {
+    IPipeReader Input { get; }
+    IPipeWriter Output { get; }
+}
+
+public struct FlushResult {
```
jaredpar: `readonly struct`

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
```diff
@@ -0,0 +1,69 @@
...
+    public bool IsCancelled { get; }
+    public bool IsCompleted { get; }
+}
+public struct ReadResult {
```
jaredpar: `readonly struct`

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
```diff
@@ -0,0 +1,69 @@
...
+    public ReadResult(ReadOnlyBuffer buffer, bool isCancelled, bool isCompleted);
+    public ReadOnlyBuffer Buffer { get; }
+    public bool IsCancelled { get; }
+    public bool IsCompleted { get; }
+}
+public class Pipe : IAwaiter<FlushResult>, IAwaiter<ReadResult>, IOutput, IPipe, IPipeReader, IPipeWriter {
```
jaredpar: sealed?

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
```diff
@@ -0,0 +1,69 @@
...
+    public Pipe(PipeOptions options);
+    public IPipeReader Reader { get; }
+    public IPipeWriter Writer { get; }
+    public void Reset();
+}
+
+public class PipeOptions {
```
jaredpar: sealed?

KrzysztofCwalina:
@benaadams, said: 
>Change ReadResult to:
```c#
[Flags]
public enum ReadStatus : byte
{
    IsCancelled,
    IsCompleted,
    IsFlush
}

public struct ReadResult {
    public ReadResult(ReadOnlyBuffer buffer, bool isCancelled, bool isCompleted);
    public ReadOnlyBuffer Buffer { get; }
    public ReadStatus Status { get; }
}
```

What about:
```c#
public enum AsyncStatus : byte
{
    IsCancelled,
    IsCompleted,
    IsCommited
}
public struct AsyncResult<T> {
    public AsyncStatus Status { get; }
    public T Result { get; }
}
```

davidfowl:
@KrzysztofCwalina over engineered and the nested generics are a smell. We also don't need the enum. I'm not sure what IsCommitted means.

halter73:
IsCommitted is just an alternative name for @benaadams's IsFlush.

KrzysztofCwalina:
@davidfowl, but wouldn't it be nice if pipelines (byte channels) were just 5 new types?

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
```diff
@@ -0,0 +1,69 @@
...
+public interface IPipeWriter : IOutput {
+    ValueAwaiter<FlushResult> FlushAsync(CancellationToken cancellationToken=default(CancellationToken));
+    void Commit();
+
+    void CancelPendingFlush();
+    void Complete(Exception exception=null);
+    void OnReaderCompleted(Action<Exception, object> callback, object state);
+}
+
+public interface IPipeConnection : IDisposable {
```
KrzysztofCwalina: We decided to call it IDuplexPipe

benaadams:
Terrible name; but what I was trying to convey is:
```csharp
[Flags]
public enum ReadStatus : byte
{
    IsCancelled,
    IsMoreInputData, // Not-flush
    IsComplete 
}
```

omariom:
@benaadams `ShipIt!` :smile:

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
pakrym: Let's clarify this in Reader/Writer review.

2018/System.IO.Pipelines_2/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Decisions
+
+* Rename `Position` to `SequenceIndex`
+* Rename `IPipeConnection` to `IDuplexPipe`
+* Make `Pipe` non-abstract
+    - Add `Reset` method
+    - Has an internal ctor
+    - Customers building their own pipes should create their own type
+* Rename `Scheduler` to `PipeScheduler` and move to `System.IO.Pipelines`
```
pakrym: Should it be an interface?

2018/System.IO.Pipelines_2/README.md
```diff
@@ -0,0 +1,104 @@
...
+    - Offer an constructor with notarguments that passes in
+      `PipeOptions.Default`
+* `IOutput`
+    - In order to settle on `IOutput` as the key constraint for writers, we need
+      a way get a `Span<T>` as this is often more efficient.
+    - We could split into `IOutput` into `ISpanOutput` `IMemoryOutput`
+
+## Decisions
+
+* Rename `Position` to `SequenceIndex`
```
KrzysztofCwalina: @pakrym was trying to use SequenceIndex and it has problems because now properties would have to be called Index which in turn clashes with int based Index properties. We decided to call this type SequencePosition.

2018/System.IO.Pipelines_2/README.md
terrajobst: Fixed

2018/System.IO.Pipelines_2/README.md
halter73: What do you mean it clashes with int based index properties? Do you expect `SequenceIndex index` and `int index` to be parameters to the same method? I'm pretty sure SequenceIndex is what we agreed to not that I'm overly attached to it.

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
terrajobst: Added to the decisions

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
terrajobst: Done

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
terrajobst: Done

2018/System.IO.Pipelines_2/README.md
halter73: Oh. Properties not parameters. What types have int and SequenceIndex indices?

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
terrajobst: Fixed

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
terrajobst: Fixed

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
terrajobst: Fixed

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
terrajobst: We don't seal unless necessary

2018/System.IO.Pipelines_2/System.IO.Pipelines2.md
terrajobst: Same

2018/System.IO.Pipelines_2/README.md
terrajobst: Let's close on this Monday with Toub

terrajobst:
I'm going to merge the status quo; for new decisions I'll create a new PR. In the ends, these are meant to capture notes, not final specs.

tmds:
When the Transport is reading the data to put it on the network, the `IsMoreInputData == true` case will be ignored by the Transport (in order to reduce syscalls). It seems wasteful to schedule the read completion (possibly deferring it to the Transport thread), if the reader is going to ignore it.

benaadams:
> When the Transport is reading the data to put it on the network, the `IsMoreInputData == true` case will be ignored by the Transport (in order to reduce syscalls). It seems wasteful to schedule the read completion (possibly deferring it to the Transport thread), if the reader is going to ignore it.

The Transport *is* the reader in this example; so it pays attention the the flag to reduce syscalls. The pipeline schedule is user mode; whereas the network write is kernel mode. So it seems a desirable trade off to increase the former to be able decrease the latter. Although in practice its not changing the pipeline scheduling; just passing additional information.

e.g. given the [Kestrel LargeResponseApp sample](https://github.com/aspnet/KestrelHttpServer/blob/bbe8851f32984264282d80ea1a0fe4c70cbffa95/samples/LargeResponseApp/Startup.cs)
```csharp
var body = context.Response.Body;
for (int i = 0; i < numChunks; i++)
{
    await body.WriteAsync(_chunk, 0, _chunkSize);
}
```
With the current implementation this is conceptually the same as
```csharp
var body = context.Response.Body;
for (int i = 0; i < numChunks; i++)
{
    await body.WriteAsync(_chunk, 0, _chunkSize);
    await body.FlushAsync();
}
```
I'm suggesting it would be better to have the ability change the behaviour to:
```csharp
var body = context.Response.Body;
for (int i = 0; i < numChunks; i++)
{
    await body.WriteAsync(_chunk, 0, _chunkSize);
}

await body.FlushAsync();
```

tmds:
> So it seems a desirable trade off to increase the former to be able decrease the latter.
Although in practice its not increasing the pipeline scheduling; just passing additional information.

Definitely.

Waking up the reader still is a bit of a waste in this case. To avoid it, the pipe could buffer up to some amount before it starts waking up the reader. It is an optimization that could be implemented later (I don't think it affects the reader-writer API).

I wonder if WriteAsync shouldn't call FlushAsync occasionally depending on the amount of data in the Pipe. Otherwise the reader may keep ignoring it (i.e. waiting till `IsMoreInputData == false`), causing too much data to be buffered in the Pipe.

benaadams:
>  To avoid it, the pipe could buffer up to some amount before it starts waking up the reader.

Already there, its what the sync non-value returning `void Commit()` does.

> I wonder if WriteAsync shouldn't call FlushAsync occasionally depending on the amount of data in the Pipe. 

WriteAsync is currently just a convience extenstion method that  [**always** calls FlushAsync](https://github.com/dotnet/corefxlab/blob/b5b8dab751fe0c5f3f8dbf381cbe87e541f17a20/src/System.IO.Pipelines/System/IO/Pipelines/PipeExtensions.cs#L14-L32); so it always says "make a syscall now"; which is the issue I have.

tmds:
> WriteAsync is currently just a convience extenstion method that always calls FlushAsync; so it always says "make a syscall now"; which is the issue I have.

You are proposing it shouldn't always call FlushAsync, right? I understand the rationale for that.
My question is: if WriteAsync doesn't _always_ call FlushAsync, shouldn't it be calling it _occasionally_ to avoid too much data being buffered in the Pipe?

omariom:
May be Pipes should not have such high level api as WriteAsync at all.

benaadams:
> My question is: if WriteAsync doesn't always call FlushAsync, shouldn't it be calling it occasionally to avoid too much data being buffered in the Pipe?

It should throw an exception if it returns continue to the writer its over the high water mark (may already do this)

tmds:
> May be Pipes should not have such high level api as WriteAsync at all.

I think this is used to implement the Stream.WriteAsync on top of a PipeWriter.

> It should throw an exception if it returns continue to the writer its over the high water mark (may already do this)

Wouldn't it be nicer to make WriteAsync do a FlushAsync at that point instead?

benaadams:
> Wouldn't it be nicer to make WriteAsync do a FlushAsync at that point instead?

Perhaps; then it would be automagic, and throw exception if FlushAsync returns continue without decreasing buffer?

tmds:
> Perhaps; then it would be automagic (and changeable via configuration), and throw exception if FlushAsync returns continue without decreasing buffer?

Yes, it buffers some amount and when it reaches the threshold it flushes (like BufferedStream). I don't understand the semantics enough to say whether it should flush at the low or high threshold (or yet another threshold).

tmds:
Suppose WriteAsync behaved this way (flushing at a threshold). Perhaps in the TLS case, this buffer would be set to 16k? Then it would no longer be necessary to make ReadAsync complete for each WriteAsync.
The ReadResult would then be either an explicit flush (writer called FlushAsync) or an implicit flush (writer called ReadAsync and threshold was reached).
When the TLS reader sees an implicit flush which is longer than 16k, it would leave the remainder since he knows more data is expected.

tmds:
> I don't understand the semantics enough to say whether it should flush at the low or high threshold (or yet another threshold).

I think buffering till the low threshold would work fine.

PR #60, opened 2018-01-24
Add 2018/System.IO.Enumerations/

terrajobst:
@JeremyKuhne please take a look to make sure I captured everything correctly.

### DIFF
--- a/2018/System.IO.Enumerations/README.md
+++ b/2018/System.IO.Enumerations/README.md
@@ -1,0 +1,20 @@
+# System.IO.Enumerations
+
+Status: **Needs more work** | 
+[API](System.IO.FileSystem.md) |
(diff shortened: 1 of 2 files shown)


cdmihai:
My feedback:
- required for CPS and benefit for MSBuild: BeginDirectoryEnumeration / EndDirectoryEnumeration callbacks.
- benefit for MSBuild: enumeration overload accepting a RecurseDirectory callback that also returns the new state object to use for that subtree (either this or we'll have to use a global dictionary in the global state object to keep the state per subtree).

terrajobst:
@cdmihai 

Thanks, added to the notes

ahsonkhan:
> required for CPS and benefit for MSBuild

What is CPS?

JeremyKuhne:
> What is CPS?

Common Project System. One of the components of VS.

PR #61, opened 2018-01-31
Second pass of System.IO.Enumeration

terrajobst:
@JeremyKuhne PTAL

### DIFF
--- a/2018/System.IO.Enumerations2/README.md
+++ b/2018/System.IO.Enumerations2/README.md
@@ -1,0 +1,15 @@
+# System.IO.Enumerations
+
+Status: **Approved with Feedback** | 
+[API](System.IO.FileSystem.md)
+
+## Notes
+
+- `EnumerationOptions`
+    - Check whether we can use a different word than `Inaccessible`
+    - `Recurse` should be `EnumerateSubdirectories`
+    - `Default` and `default` don't necessarily have the same values, which
+      might be confusing.
+    - We should make it a class, it solves most problems. And for allocations,
+      it should be a non-issue as the options themselves are static and don't
+      need to be created per enumeration.
--- a/2018/System.IO.Enumerations2/README.md
+++ b/2018/System.IO.Enumerations2/README.md
@@ -1,7 +1,7 @@
 # System.IO.Enumerations
 
 Status: **Approved with Feedback** | 
-[API](System.IO.FileSystem.md)
+[API](System.IO.Enumeration.md)
 
 ## Notes
 
--- a/2018/System.IO.Enumerations2/System.IO.Enumeration.md
+++ b/2018/System.IO.Enumerations2/System.IO.Enumeration.md
@@ -218,7 +218,7 @@
 +   public MatchType MatchType { get; set; }
 +   public bool Recurse { get; set; }
   }
-+ public struct FileSystemEntry {
++ public ref struct FileSystemEntry {
 +   public FileAttributes Attributes { get; }
 +   public DateTimeOffset CreationTimeUtc { get; }
 +   public ReadOnlySpan<char> Directory { get; }
--- a/2018/System.IO.Enumerations2/System.IO.Enumeration.md
+++ b/2018/System.IO.Enumerations2/System.IO.Enumeration.md
@@ -1,0 +1,273 @@
+# System.IO.FileSystem
+
+```diff
+---System.IO.FileSystem.dll (.NET Core 2.0)
++++System.IO.FileSystem.dll (.NET Core 2.1)
+ namespace System.IO {
+  public static class Directory {
+    public static DirectoryInfo CreateDirectory(string path);
+    public static void Delete(string path);
+    public static void Delete(string path, bool recursive);
+    public static IEnumerable<string> EnumerateDirectories(string path);
+    public static IEnumerable<string> EnumerateDirectories(string path, string searchPattern);
++   public static IEnumerable<string> EnumerateDirectories(string path, string searchPattern, EnumerationOptions enumerationOptions);
+    public static IEnumerable<string> EnumerateDirectories(string path, string searchPattern, SearchOption searchOption);
+    public static IEnumerable<string> EnumerateFiles(string path);
+    public static IEnumerable<string> EnumerateFiles(string path, string searchPattern);
++   public static IEnumerable<string> EnumerateFiles(string path, string searchPattern, EnumerationOptions enumerationOptions);
+    public static IEnumerable<string> EnumerateFiles(string path, string searchPattern, SearchOption searchOption);
+    public static IEnumerable<string> EnumerateFileSystemEntries(string path);
+    public static IEnumerable<string> EnumerateFileSystemEntries(string path, string searchPattern);
++   public static IEnumerable<string> EnumerateFileSystemEntries(string path, string searchPattern, EnumerationOptions enumerationOptions);
+    public static IEnumerable<string> EnumerateFileSystemEntries(string path, string searchPattern, SearchOption searchOption);
+    public static bool Exists(string path);
+    public static DateTime GetCreationTime(string path);
+    public static DateTime GetCreationTimeUtc(string path);
+    public static string GetCurrentDirectory();
+    public static string[] GetDirectories(string path);
+    public static string[] GetDirectories(string path, string searchPattern);
++   public static string[] GetDirectories(string path, string searchPattern, EnumerationOptions enumerationOptions);
+    public static string[] GetDirectories(string path, string searchPattern, SearchOption searchOption);
+    public static string GetDirectoryRoot(string path);
+    public static string[] GetFiles(string path);
+    public static string[] GetFiles(string path, string searchPattern);
++   public static string[] GetFiles(string path, string searchPattern, EnumerationOptions enumerationOptions);
+    public static string[] GetFiles(string path, string searchPattern, SearchOption searchOption);
+    public static string[] GetFileSystemEntries(string path);
+    public static string[] GetFileSystemEntries(string path, string searchPattern);
++   public static string[] GetFileSystemEntries(string path, string searchPattern, EnumerationOptions enumerationOptions);
+    public static string[] GetFileSystemEntries(string path, string searchPattern, SearchOption searchOption);
+    public static DateTime GetLastAccessTime(string path);
+    public static DateTime GetLastAccessTimeUtc(string path);
+    public static DateTime GetLastWriteTime(string path);
+    public static DateTime GetLastWriteTimeUtc(string path);
+    public static string[] GetLogicalDrives();
+    public static DirectoryInfo GetParent(string path);
+    public static void Move(string sourceDirName, string destDirName);
+    public static void SetCreationTime(string path, DateTime creationTime);
+    public static void SetCreationTimeUtc(string path, DateTime creationTimeUtc);
+    public static void SetCurrentDirectory(string path);
+    public static void SetLastAccessTime(string path, DateTime lastAccessTime);
+    public static void SetLastAccessTimeUtc(string path, DateTime lastAccessTimeUtc);
+    public static void SetLastWriteTime(string path, DateTime lastWriteTime);
+    public static void SetLastWriteTimeUtc(string path, DateTime lastWriteTimeUtc);
+  }
+  public sealed class DirectoryInfo : FileSystemInfo {
+    public DirectoryInfo(string path);
+    public override bool Exists { get; }
+    public override string Name { get; }
+    public DirectoryInfo Parent { get; }
+    public DirectoryInfo Root { get; }
+    public void Create();
+    public DirectoryInfo CreateSubdirectory(string path);
+    public override void Delete();
+    public void Delete(bool recursive);
+    public IEnumerable<DirectoryInfo> EnumerateDirectories();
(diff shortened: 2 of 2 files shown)


2018/System.IO.Enumerations2/System.IO.Enumeration.md
```diff
@@ -0,0 +1,273 @@
...
++   public long Length { get; }
++   public string OriginalRootDirectory { get; }
++   public string RootDirectory { get; }
++   public DirectoryInfo ToDirectoryInfo();
++   public FileInfo ToFileInfo();
++   public FileSystemInfo ToFileSystemInfo();
++   public string ToFullPath();
++   public string ToSpecifiedFullPath();
+  }
++ public class FileSystemEnumerable<TResult, TState> : IEnumerable, IEnumerable<TResult> {
```
jkotas: Why do we need the state? Can't folks just inherit from this if they need to carry state around?

terrajobst:
By the way, @JeremyKuhne talked with @VSadov about the struct size. The conclusion was that the compiler only has to make copies when the struct itself is returned from read-only positions or when inside a try block.

The former is impossible (because `FileSystemEntry` will be a `ref` struct and also be passed by `ref`, so it's always known to be non-readonly). The only potential gotcha is a try block, but consumers aren't likely to have one. And if they do and care about perf, they'll measure :-)

2018/System.IO.Enumerations2/System.IO.Enumeration.md
```diff
@@ -0,0 +1,273 @@
...
++   public int BufferSize { get; set; }
++   public static EnumerationOptions Default { get; }
++   public bool IgnoreInaccessible { get; set; }
++   public MatchType MatchType { get; set; }
++   public bool Recurse { get; set; }
+  }
++ public struct FileSystemEntry {
++   public FileAttributes Attributes { get; }
++   public DateTimeOffset CreationTimeUtc { get; }
++   public ReadOnlySpan<char> Directory { get; }
```
jkotas: Is the returned Span immutable, or is it going to change under me when I get the next entry?

JeremyKuhne: FileSystemEntry is passed by ref, nothing in it is valid after the call finishes.

2018/System.IO.Enumerations2/README.md
```diff
@@ -0,0 +1,15 @@
+# System.IO.Enumerations
+
+Status: **Approved with Feedback** | 
+[API](System.IO.FileSystem.md)
```
jkotas: This link is broken.

2018/System.IO.Enumerations2/System.IO.Enumeration.md
JeremyKuhne: I was actually thinking of including both options. Both scenarios are pretty common (state/no state)

``` c#
    public class FileSystemEnumerable<TResult> : Collections.Generic.IEnumerable<TResult>
    {
        public FileSystemEnumerable(string directory, FindTransform transform, EnumerationOptions options = null) { }

        public FindPredicate ShouldRecursePredicate { get { throw null; } set { } }
        public FindPredicate ShouldIncludePredicate { get { throw null; } set { } }

        public Collections.Generic.IEnumerator<TResult> GetEnumerator() { throw null; }
        Collections.IEnumerator Collections.IEnumerable.GetEnumerator() { throw null; }


        public delegate bool FindPredicate(ref FileSystemEntry entry);
        public delegate TResult FindTransform(ref FileSystemEntry entry);
    }
    public class FileSystemEnumerable<TResult, TState> : Collections.Generic.IEnumerable<TResult>
    {
        public FileSystemEnumerable(string directory, FindTransform transform, EnumerationOptions options = null) { }

        public FindPredicate ShouldRecursePredicate { get { throw null; } set { } }
        public FindPredicate ShouldIncludePredicate { get { throw null; } set { } }
        public TState State { get { throw null; } set { } }

        public Collections.Generic.IEnumerator<TResult> GetEnumerator() { throw null; }
        Collections.IEnumerator Collections.IEnumerable.GetEnumerator() { throw null; }

        public delegate bool FindPredicate(ref FileSystemEntry entry, TState state);
        public delegate TResult FindTransform(ref FileSystemEntry entry, TState state);
    }
```

2018/System.IO.Enumerations2/System.IO.Enumeration.md
```diff
@@ -0,0 +1,273 @@
...
+  }
+ }
++namespace System.IO.Enumeration {
++ public struct EnumerationOptions {
++   public FileAttributes AttributesToSkip { get; set; }
++   public int BufferSize { get; set; }
++   public static EnumerationOptions Default { get; }
++   public bool IgnoreInaccessible { get; set; }
++   public MatchType MatchType { get; set; }
++   public bool Recurse { get; set; }
```
JeremyKuhne: Currently have this as `RecurseSubdirectories` based on feedback

2018/System.IO.Enumerations2/System.IO.Enumeration.md
jkotas: But why does the state need to be generic argument? Since this is a class, inheriting from it is more footprint friendly.

2018/System.IO.Enumerations2/System.IO.Enumeration.md
```diff
@@ -0,0 +1,273 @@
...
+    public virtual void GetObjectData(SerializationInfo info, StreamingContext context);
+    public void Refresh();
+  }
+  public enum SearchOption {
+    AllDirectories = 1,
+    TopDirectoryOnly = 0,
+  }
+ }
++namespace System.IO.Enumeration {
++ public struct EnumerationOptions {
```
JeremyKuhne: Writing tests for this it is a pain to keep filtering out "." and "..". Since it is a rare case, adding the option

``` c#
        public bool ReturnSpecialDirectories { get { throw null; } set { } }
```

makes a lot of sense.

2018/System.IO.Enumerations2/System.IO.Enumeration.md
jkotas: And what I store the FileSystemEntry somewhere, like in static?

```
static FileSystemEntry s_cached;

    s_cached = ... entry;
...

File: 2018/System.IO.Enumerations2/System.IO.Enumeration.md (hunk 1 above)
Comment (JeremyKuhne) [reply 2018/System.IO.Enumerations2/System.IO.Enumeration.md]: > But why does the state need to be generic argument? Since this is a class, inheriting from it is more footprint friendly.

Is it? There would only be one for each type of state people need, right? Wouldn't that potentially be less if a number of folks derive with a string state?

File: 2018/System.IO.Enumerations2/System.IO.Enumeration.md (hunk 2 above)
Comment (JeremyKuhne) [reply 2018/System.IO.Enumerations2/System.IO.Enumeration.md]: You shouldn't be able to as it is a ref struct. Which isn't described here.

File: 2018/System.IO.Enumerations2/System.IO.Enumeration.md [hunk 6]
```diff
@@ -0,0 +1,273 @@
...
++   public FileAttributes AttributesToSkip { get; set; }
++   public int BufferSize { get; set; }
++   public static EnumerationOptions Default { get; }
++   public bool IgnoreInaccessible { get; set; }
++   public MatchType MatchType { get; set; }
++   public bool Recurse { get; set; }
+  }
++ public struct FileSystemEntry {
```
JeremyKuhne: This should be `public ref struct`

jkotas:
> `FileSystemEntry` will be a ref struct and also be passed by `ref`

I do not see it marked as `ref` in the attached md file nor in the feedback.

JeremyKuhne:
> I do not see it marked as ref in the attached md file nor in the feedback.

It got lost in translation somewhere. I've added a comment that it is missing.

2018/System.IO.Enumerations2/System.IO.Enumeration.md
jkotas: I expect that the state would be typically a struct. It is what people do for these TState arguments.

There is a copy&paste of all the code in your generic type for each struct instantiation at runtime...

If I am reading this right, the TState is just passed to a delegate. Can't the state be just associated with the delegate?

JeremyKuhne: > There is a copy&paste of all the code in your generic type for each struct instantiation at runtime...

Didn't know that, thanks!

> If I am reading this right, the TState is just passed to a delegate. Can't the state be just associated with the delegate?

You mean via capturing? I was under the impression that capturing would prevent the delegate from being cached. Am I misunderstanding?

jkotas: Yes, the delegate would not be cached. But this is going to allocate quite a bit anyway, so allocating the delegate as well is not a big deal.

JeremyKuhne: > allocating the delegate as well is not a big deal

Just trying to keep the initial allocations to a minimum, but totally willing to add some to avoid horrible contortions. The key value in this API is low allocations per MoveNext().  I suppose one can always derive to cut a few more initial allocs as well.

2018/System.IO.Enumerations2/System.IO.Enumeration.md
danmosemsft: That name requires me consult the documentation to see whether I want it. I suggest something like ReturnDotDotDirectories or something that makes clear that almost surely you don't want this.

Did we conclude that there was  a scenario where you want the dotdots?

VSadov:
The part about implicit copying.

Compiler copies structs when invoking a method on a struct and:
- struct is not a `readonly` struct (then we know that methods are not mutating)
- the receiver of a call is a `readonly` variable (then we must make a "defensive copy" of the receiver)

```C#
readonly S1 roField;

void M1(in S1 inParam, ref S1 refParam)
{
      roField.Method();     // invoked on a copy since receiver is readonly
      inParam.Method();   // invoked on a copy since receiver is readonly
      refParam.Method(); // no copy, mutations from Method can and will propagate to refParam
}
```

2018/System.IO.Enumerations2/System.IO.Enumeration.md
JeremyKuhne: They aren't always returned from the OS so we need to at least have the option. I'm not set on any particular name here, maybe a few others can chime in?

JeremyKuhne:
Thanks for the clarification @VSadov!

For those that are curious, our struct cannot be made readonly as the Unix implementation mutates, lazily filling in the data that isn't part of what the OS gives us (time stamps, UTF-16, etc.).

VSadov:
The thing about "in a `try` block" :

Compiler tries to avoid copies when struct instances are constructed. Most common case like
```C#
S1 x = new S1(arg1, arg2);
```
are actually emitted as IL version of

```C#
// inplace reinitialization
x..ctor(arg1, arg2);
```
There are cases when optimization above would be observable - if there is a chance that `.ctor` throws after initializing half the fields.
The language spec says that assignment of the whole thing happens _after_ `.ctor` runs successfully, so seeing partially updated struct would be in violation of the spec (and common sense). 

So, if we believe that there is a chance that you could see partial result, we will not perform such optimization and will emit code that literally constructs an instance and then copies/assigns it to the provided local. JIT will have its chance at eliding the copy, but I have doubts it often can. Especially when structs are large.

The reasons for such unoptimized copy are:
- when initialization of `x` is inside `try`. The reasoning is that when `.ctor`. throws without finishing, the code in handlers will have a chance to observe badly initialized instance. We do not look into what handlers actually do, we just assume they can do that.
- variable is not local to the method. I.E.  It is a ref parameter or a field on a reference type or something like that that may outlive the method. 
So even if your method crashes, someone up the call stack might have an exception handler where it may have access to the variable you are initializing.

2018/System.IO.Enumerations2/README.md
terrajobst: Fixed

terrajobst:
@jkotas @JeremyKuhne 

Yeah, API reviewer doesn't handle ref structs apparently.

2018/System.IO.Enumerations2/System.IO.Enumeration.md
terrajobst: Fixed

2018/System.IO.Enumerations2/System.IO.Enumeration.md
JeremyKuhne: @terrajobst Can we change to:

``` c#
    public class FileSystemEnumerable<TResult> : Collections.Generic.IEnumerable<TResult>
    {
        public FileSystemEnumerable(string directory, FindTransform transform, EnumerationOptions options = null) { }

        public FindPredicate ShouldRecursePredicate { get { throw null; } set { } }
        public FindPredicate ShouldIncludePredicate { get { throw null; } set { } }

        public Collections.Generic.IEnumerator<TResult> GetEnumerator() { throw null; }
        Collections.IEnumerator Collections.IEnumerable.GetEnumerator() { throw null; }


        public delegate bool FindPredicate(ref FileSystemEntry entry);
        public delegate TResult FindTransform(ref FileSystemEntry entry);
    }
```

We can always add one with state later if we find a stronger call for it.

PR #62, opened 2018-03-16
WIP: Add .NET Core 2.0 vs. .NET Core 2.1

terrajobst:
@KrzysztofCwalina @joshfree @ahsonkhan

This isn't ready yet -- as discussed in the meeting I'll add summarize and assign the area owners.

PR #63, opened 2018-03-22
API diff between .NET Core 2.0 and .NET Core 2.1

terrajobst:
This is meant as a final pass to make sure we haven't missed anything.

- [x] Merge comments from #62 
- [x] Assign area owners

### DIFF
--- a/2018/NetCore20-vs-NetCore21-Diff/README.md
+++ b/2018/NetCore20-vs-NetCore21-Diff/README.md
@@ -1,0 +1,1893 @@
+# API diff between .NET Core 2.0 and .NET Core 2.1
+
+To the best of my knowledge this is an accureate diff between .NET Core 2.0 and
+.NET Core 2.1, excluding ASP.NET Core as well as the Windows Compaptibility
+pack.
+
+Sources:
+
+* **.NET Core 2.0**. `\\fxcore\platforms\ApiCatalog\.NET Core 2.0`. This folder
+  was created based on the `/reference` arguments passed to `csc.exe` when
+  building a .NET Core 2.0 console app.
+* **.NET Core 2.1**. `\\cpvsbuild\drops\DotNetCore\DotNet-CoreFx-APICatalog-LayoutDrop\20180322-01\ApiCatalog\ref\microsoft.netcore.app`
+  This is our live build drop for .NET Core.
+
+## Action Items
+
+```diff
+ namespace System {
+-    public class ArgumentException : SystemException, ISerializable {
++    public class ArgumentException : SystemException {
+     }
+-    public class ArgumentOutOfRangeException : ArgumentException, ISerializable {
++    public class ArgumentOutOfRangeException : ArgumentException {
+     }
+     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+     public struct ArraySegment<T> : ICollection<T>, IEnumerable, IEnumerable<T>, IList<T>, IReadOnlyCollection<T>, IReadOnlyList<T> {
+-        void System.Collections.Generic.ICollection<T>.CopyTo(T[] array, int arrayIndex);
+     }
+     public static class BitConverter {
++        public static bool ToBoolean(ReadOnlySpan<byte> value);
++        public static char ToChar(ReadOnlySpan<byte> value);
++        public static double ToDouble(ReadOnlySpan<byte> value);
++        public static short ToInt16(ReadOnlySpan<byte> value);
++        public static int ToInt32(ReadOnlySpan<byte> value);
++        public static long ToInt64(ReadOnlySpan<byte> value);
++        public static float ToSingle(ReadOnlySpan<byte> value);
++        public static ushort ToUInt16(ReadOnlySpan<byte> value);
++        public static uint ToUInt32(ReadOnlySpan<byte> value);
++        public static ulong ToUInt64(ReadOnlySpan<byte> value);
++        public static bool TryWriteBytes(Span<byte> destination, char value);
++        public static bool TryWriteBytes(Span<byte> destination, short value);
++        public static bool TryWriteBytes(Span<byte> destination, int value);
++        public static bool TryWriteBytes(Span<byte> destination, long value);
++        public static bool TryWriteBytes(Span<byte> destination, ushort value);
++        public static bool TryWriteBytes(Span<byte> destination, uint value);
++        public static bool TryWriteBytes(Span<byte> destination, ulong value);
++        public static bool TryWriteBytes(Span<byte> destination, float value);
++        public static bool TryWriteBytes(Span<byte> destination, double value);
++        public static bool TryWriteBytes(Span<byte> destination, bool value);
+     }
+     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+     public struct Boolean : IComparable, IComparable<bool>, IConvertible, IEquatable<bool> {
++        public static Boolean Parse(ReadOnlySpan<char> value);
++        public Boolean TryFormat(Span<char> destination, out int charsWritten);
++        public static Boolean TryParse(ReadOnlySpan<char> value, out Boolean result);
+     }
+     [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
+     public struct Byte : IComparable, IComparable<byte>, IConvertible, IEquatable<byte>, IFormattable {
++        public static Byte Parse(ReadOnlySpan<char> s, NumberStyles style=(NumberStyles)(7), IFormatProvider provider=null);
++        public bool TryFormat(Span<char> destination, out int charsWritten, ReadOnlySpan<char> format=default(ReadOnlySpan<char>), IFormatProvider provider=null);
++        public static bool TryParse(ReadOnlySpan<char> s, out Byte result);
++        public static bool TryParse(ReadOnlySpan<char> s, NumberStyles style, IFormatProvider provider, out Byte result);
+     }
+-    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential, Size=1)]
+-    public struct ConsoleKeyInfo {
++    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
++    public struct ConsoleKeyInfo {
+     }
+     public static class Convert {
++        public static string ToBase64String(ReadOnlySpan<byte> bytes, Base64FormattingOptions options=(Base64FormattingOptions)(0));
++        public static bool TryFromBase64Chars(ReadOnlySpan<char> chars, Span<byte> bytes, out int bytesWritten);
++        public static bool TryFromBase64String(string s, Span<byte> bytes, out int bytesWritten);
++        public static bool TryToBase64Chars(ReadOnlySpan<byte> bytes, Span<char> chars, out int charsWritten, Base64FormattingOptions options=(Base64FormattingOptions)(0));
+     }
+-    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential, Size=1)]
+-    public struct DateTime : IComparable, IComparable<DateTime>, IConvertible, IEquatable<DateTime>, IFormattable, ISerializable {
++    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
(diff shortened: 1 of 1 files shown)


2018/NetCore20-vs-NetCore21-Diff/README.md
```diff
@@ -0,0 +1,1893 @@
+# API diff between .NET Core 2.0 and .NET Core 2.1
+
+To the best of my knowledge this is an accureate diff between .NET Core 2.0 and
```
jnm2: Typo `accureate`

2018/NetCore20-vs-NetCore21-Diff/README.md
```diff
@@ -0,0 +1,1893 @@
...
+.NET Core 2.1, excluding ASP.NET Core as well as the Windows Compaptibility
```
jnm2: Typos `accureate` and `Compaptibility`

terrajobst:
> +        bool System.IEquatable<System.IntPtr>.Equals(IntPtr value);

Why did we change it?

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:179 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +        public static implicit operator Memory<T> (ArraySegment<T> arraySegment);

should be `segment`

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:205 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +        public static implicit operator ReadOnlyMemory<T> (ArraySegment<T> arraySegment);

should be `segment`

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:320 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +        public static implicit operator ReadOnlySpan<T> (ArraySegment<T> arraySegment);

should be `segment`

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:348 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +        public static implicit operator Span<T> (ArraySegment<T> arraySegment);

should be `segment`

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:408 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +        protected internal abstract bool TryGetArray(out ArraySegment<T> arraySegment);

should be `segment`

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:568 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +        public static implicit operator Memory<T> (ArraySegment<T> arraySegment);

should be `segment`

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:205 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +        bool System.IEquatable<System.UIntPtr>.Equals(UIntPtr value);

`other`

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:482 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
>  namespace System.Buffers {

Should we revisit the contents of this namespace? It seems we have a bunch of types that are plumbing that should be in a nested namespace, such as `IRetainable`.

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:519 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +    public delegate void ReadOnlySpanAction<T, in TArg>(ReadOnlySpan<T> span, TArg arg); {

Looks like plumbing. Should be in a sub namespace.

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:612 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +    public delegate void SpanAction<T, in TArg>(Span<T> span, TArg arg); {

Looks like plumbing. Should be in a sub namespace.

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:618 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +        public static bool TryWriteInt16BigEndian(Span<byte> destination, short value);

All `TryWriteXxx` should use `destination`

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:680 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +        public static void ZeroMemory(Span<byte> buffer);

`destination`

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:1663 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
> +        public static bool FixedTimeEquals(ReadOnlySpan<byte> left, ReadOnlySpan<byte> right);

Should the parameters be named `span`/`other` or `left`/`right`? These are statics hence I'm leaning towards the latter.

---
Refers to: 2018/NetCore20-vs-NetCore21-Diff/README.md:1662 in 9b5c027. [](commit_id = 9b5c0274be056047231fbf5dad321e85708196cc, deletion_comment = False)

terrajobst:
@stephentoub 

> > Looks like plumbing. Should be in sub namespace.
>
> @terrajobst, can you elaborate?  These are essentially just `Action<..>`, but because span can't be used as a generic type parameter, we had to create custom delegates where we needed delegates that take span, e.g. `String.Create`.
> 
> If that's all understood and the agreement was still that they should be in a subnamespace, what namespace?

@KrzysztofCwalina brought it up that we [should do another pass](https://github.com/dotnet/apireviews/pull/63#issuecomment-375418439) on the `System.Buffers`. I understand why these delegates are needed but they are currently used by a single ctor on `string` and seem to be more advanced and thus rarely used.

danmosemsft:
Not @ahsonkhan probably @maryamariyan

2018/NetCore20-vs-NetCore21-Diff/README.md
```diff
@@ -0,0 +1,1893 @@
...
++    public static class DbProviderFactories {
++        public static DbProviderFactory GetFactory(DataRow providerRow);
++        public static DbProviderFactory GetFactory(DbConnection connection);
++        public static DbProviderFactory GetFactory(string providerInvariantName);
++        public static DataTable GetFactoryClasses();
++        public static IEnumerable<string> GetProviderInvariantNames();
++        public static void RegisterFactory(string providerInvariantName, DbProviderFactory factory);
++        public static void RegisterFactory(string providerInvariantName, string factoryTypeAssemblyQualifiedName);
++        public static void RegisterFactory(string providerInvariantName, Type providerFactoryClass);
++        public static
```
ahsonkhan: Should we consider changing this to chars and become inconsistent with the array-based Read method but consistent with the BinaryWriter Write methods?

ahsonkhan: Why were these parameters renamed? It looks like they were done as far back as https://github.com/dotnet/corefx/commit/e3fc18c58396737ee5e6bc80acf508031d63cc8e#diff-b66ed9c9f50e26cf2e4afaa99ec250b8.

These already shipped as the following in .NET Core 2.0:
```C#
public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback callback, object state);
public override IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback callback, object state);
```

cc @JeremyKuhne

ahsonkhan: Inconsistent parameter naming here.

cc @ViktorHofer, @ianhays

ahsonkhan: buffer?

ahsonkhan: inputBuffer, outputBuffer?

ahsonkhan: other? span?

Existing API is `public bool Equals(StringBuilder sb);`

JeremyKuhne: This one was a bug. We renamed from the original on desktop. "array" is the right parameter name.

JeremyKuhne: I would keep consistent with this class and not change from buffer.

bartonjs:
It's not obvious how to reply to any of the line item comments.

Crypto:

> public static void ZeroMemory(Span<byte> buffer);

We could probably do better than "buffer", but "destination" seems wrong (it's not the output of any serialization or computation.  In discussion with Krzysztof it seems like we could rename it to "span", but that's probably not enough value-add to do it at this point.

> public static bool FixedTimeEquals(ReadOnlySpan<byte> left, ReadOnlySpan<byte> right);

Left and right are what Krzysztof came up with when prompted.  I agree they're better than "span/other" since this is a static non-extension.  The best comparison is IEquatable\<T\>, which uses "x/y", but we don't like that, either.  So "left/right" still seems like the best for static equals methods.

2018/NetCore20-vs-NetCore21-Diff/README.md
bartonjs: This seems the same to me as the Brotli(etc) compress or decompress, which are `(source, destination)`.  So I think we would rename this only if we were renaming it on compression.  Or am I missing something?

maryamariyan: OK. I am assuming it can be kept as `(source, destination)` in both places.

maryamariyan: I'll change buffer in `BeginRead(byte[] buffer, ...)` to array to make consistent with the rest.

maryamariyan: based on the response below by @bartonjs, I'll keep it as is.

terrajobst: AFAICT these APIs aren't net-new. These are ports from .NET Framework. Changing them isn't an option.

terrajobst: The diff is messed up. No idea what's happening on GitHub...

terrajobst:
Apparently GitHub diffs aren't the most compelling way to track this as they comments aren't really pointing to the correct line anymore.

Besides the namespace, are there are any open issues? @maryamariyan, I assume you thumbed up the ones you took care of, right?

2018/NetCore20-vs-NetCore21-Diff/README.md
ahsonkhan: Yes, the line numbers don't line up. Maybe because `large diffs are not rendered by default`.

terrajobst: In my own code review tool they show up properly. Way to go GitHub...

terrajobst: Agreed, local consistency with members on the same existing type trumps consistency with members on other types.

terrajobst: 👍

terrajobst: I'd prefer `span`.

maryamariyan:
@terrajobst yes I thumbed up the ones I took care of and the ones with hooray emoji are meant to show no action was required as it was already fixed or not needed change

terrajobst:
> @KrzysztofCwalina brought it up that we [should do another pass](https://github.com/dotnet/apireviews/pull/63#issuecomment-375418439) on the `System.Buffers` namespace. I understand why these delegates are needed but they are currently only used by a single ctor on `string`; the question is whether they are likely becoming so popular that they should be in the feature's core namespace.

Synced with Krzysztof. We're out of time and we're happy enough that we'll leave it as-is.

2018/NetCore20-vs-NetCore21-Diff/README.md
maryamariyan: For new BrotliStream BeginRead/BeginWrite/Write/Read/ReadAsync/ReadAsync APIs I made sure we use buffer to stay consistent with Stream in PR dotnet/corefx#28581.

PR #64, opened 2018-06-20
Add notes for UTF8 strings

terrajobst:


2018/System.Utf8String/README.md
```diff
@@ -0,0 +1,57 @@
...
+      in `Utf8String`
+
+## Notes
+
+* Constructors will copy data and validate
+* Equality is ordinal (like `String`)
+* Comparisons are ordinal (unlike `String` which is culture specific)
+    - `CompareOrdinal` might be confusing as it implies the default isn't
+      ordinal. Maybe we should just drop the `Ordinal` suffix. Instead of making
+      it an overload to `Compare` we should just default the comparison
```
jkotas: Exposing the simple operation that is expected to be as fast as possible via default parameters introduces a fixed performance overhead that is hard to get rid of.

We have the same design bug in the UTF8 formatter: https://github.com/dotnet/corefx/issues/25425 . The default argument with the format specifier is hard to optimize out. It means that folks who really want to the fastest Utf8 integer formatting are still better off rolling their own.

jkotas: Also, these patterns tend to pull in a lot of code that is not used. It is a problem for tree-shaken apps. We are working on a special framework + IL linker feature to compensate for past mistakes we made here: https://github.com/dotnet/designs/pull/42.

2018/System.Utf8String/README.md
```diff
@@ -0,0 +1,57 @@
...
+      the heap (as opposed to `Span<Char8>`) but there is still some debate
+      whether that's the correct design.
+    - We should default all `StringComparison` arguments to `Ordinal`.
+    - In fact, we should use optional parameters across the board to reduce
+      number of overloads.
+* `UnicodeScalar`
+    - We probably want to change `String` and `StringBuilder` to also accept and
+      return `UnicodeScalar`
+* `Char8`
+    - Not reviewed today, but logically an 8-bit representation for a code point
```
KrzysztofCwalina: Char8 is a UTF8 code unit. It has very little to do with w "code point"

2018/System.Utf8String/README.md
```diff
@@ -0,0 +1,57 @@
...
+* `Utf8String`
+    - Immutable reference type, but we're still talking about whether it can or
+      should be a struct.
+    - Is a designed to be a user-visible exchange type that can store data to
+      the heap (as opposed to `Span<Char8>`) but there is still some debate
+      whether that's the correct design.
+    - We should default all `StringComparison` arguments to `Ordinal`.
+    - In fact, we should use optional parameters across the board to reduce
+      number of overloads.
+* `UnicodeScalar`
```
KrzysztofCwalina: It would be worth to note that we have not yet closed on the name of the type representing the scalar. UnicodeScalar is the tentative name.

2018/System.Utf8String/README.md
VSadov: If there is an obvious default, it makes sense to have a specialized overload for it. 

Reducing number of overloads by introducing optional parameters saves some metadata, but the price for that is the execution cost. 
It could be desirable if metadata savings were huge (like if there are many parameters with numerous useful permutations), but this case does not look like that.

2018/System.Utf8String/README.md
benaadams: Maybe also highlight that this would be the framework keyword like `Int32` and `Int64` which may be different from the language keyword like `int` and `long`

2018/System.Utf8String/README.md
terrajobst: Will change the wording.

2018/System.Utf8String/README.md
terrajobst: @jkotas 

> Exposing the simple operation that is expected to be as fast as possible via default parameters introduces a fixed performance overhead that is hard to get rid of.

Great point. I'll update the notes to include this.

2018/System.Utf8String/README.md
terrajobst: Based on conversations with @jaredpar it seems extremely unlikely that we'd introduce language keywords for these types as these would have to be contextual (i.e. depending on whether they can resolve to a type with the name of the keyword they'd still have to bind to this type, rather than the type represented by the keyword).

2018/System.Utf8String/README.md
KrzysztofCwalina: The point is, Char8 does not represent a "code point". Code Point range is very large and does not fit into 8 bits no matter how you slice it.

If we want it to represent code units other than UTF8 code unit, we can totally do it as long as the encoding uses single byte as its code unit, e.g. Ascii,  Latin-x, etc. The proposed name of this type does not include "Utf8" and so it could be used for other encodings. The name only says that it's text and that it's 8 bits.

terrajobst: Ah, I see. So is "8-bit code unit" then the right phrase?

terrajobst: Cool, fixed!

CyrusNajmabadi:
Not sure if you're taking comments here.  But i would definitely prefer this be a struct/readonly-struct.  It would be great if this was as low overhead as possible.  Similar to ImmutableArray, 'null-ness' could be exposed through props like `.IsDefualt` and `.IsDefaultOrEmpty`.

GrabYourPitchforks:
@CyrusNajmabadi, would using `ROM<T>` address your concern about making this as cheap as possible? Normally this concern is voiced in the context of constructing a Utf8String instance over existing data or of cheaply slicing existing data.

CyrusNajmabadi:
> @CyrusNajmabadi, would using `ROM<T>` address your concern about making this as cheap as possible?

Probably... :)

Hard t tell without seeing the API.  Thanks!

PR #66, opened 2018-10-02
System.Device.Gpio

terrajobst:


2018/System.Device.Gpio/README.md
```diff
@@ -0,0 +1,39 @@
+# System.Device.Gpio
+
+Status: **Needs more work** | 
+[API Ref](System.Device.Gpio.md) |
+[Spec](System.Device.Gpio.Proposal.md) |
+[Video](https://www.youtube.com/watch?v=OK0jDe8wtyg)
+
+## Notes
+
+* `EnableRasingEvents` looks like a usability trap. Could we do this
```
jkotas: Typo: automtaically

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
+# System.Device.Gpio Proposal
+
+The goal of this API is to allow .Net Core developers to access General Purpose
+IO (GPIO) pins of IoT devices like Raspberry Pi, Hummingboard, or Odroid. This
+will allow applications to read and write data from/to sensors, leds and other
```
jkotas: leds  -> LEDs

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+public class GpioController : IDisposable
+{
+    protected GpioDriver driver { get; }
+    public static GpioController Open(PinNumberingScheme numbering = PinNumberingScheme.Logic);
+    public static GpioController Open(GpioDriver driver, PinNumberingScheme numbering = PinNumberingScheme.Logic);
+    public static GpioController Open(GpioDriverKind kind, PinNumberingScheme numbering = PinNumberingScheme.Logic);
+    public PinNumberingScheme Numbering { get; set; }
+    public IEnumerable<IGpioConnection> OpenConnections { get; }
+    public int PinCount { get; }
+    public bool isPinOpen(int pinNumber);
```
jkotas: CamelCasing? (I guess you did not get this far during the API review...)

2018/System.Device.Gpio/README.md
aggieben: Also: `EnableRasingEvents` => `EnableRaisingEvents`?

terrajobst: Fixed

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+### Communicating with sensors using SPI and I2C
+
+*TODO: These two protocols will be discussed in the next iteration of the Api
+Review*
+
+## Proposed API
+
+Following are the proposed APIs.
+
+### GpioController class
```
janvorli: I am missing a way to read / write multiple bits of a GPIO port at once - the GPIO devices usually consist of ports with certain number of bits (most often 8) and accessing the port as a whole is useful in case when it is being used e.g. for parallel data transfer.

joperezr: I'm not entirely sure I get what you are saying, but are you refering to more complex Gpio sensors that use either the I2C or SPI protocols? For a single GpioPin, you can only really read either a 1 or a 0, there are no more bits to read/write.

It would help if you have an example of what you mean.

2018/System.Device.Gpio/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+* `EnableRaisingEvents` looks like a usability trap. Could we do this
+  automatically, e.g. when a callback is registered, when `NotifyFilter` is set,
+  or when `WaitForEvent` is called
+* Should we also offer `WaitForEventAsync`?
+* `WaitForEvent` has race condition, unless the system buffers events.
+* `GpiController.Open()` probably should have validation to ensure folks don't
+  pick the wrong board by accident.
+* Instead of `GpiController.Open()` with a enum and a drive, should we just
+  expose concrete controllers, such as `RaspberryPiController`?
```
joperezr: I don't really remember this part from the review. In fact, on the initial proposal yesterday I had specific controllers where you didn't have to pass in the driver, but feedback suggested that having instead just one controller class would be better. I think that what was mentioned on the review that we should just leave the GpioController Open method with Enum out, and instead have the user pass in a driver object (in some form) when getting a new controller

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
janvorli: I am not referring to I2C or SPI. I am referring to parallel output of data to the whole port. GPIO pins are grouped to ports that can be read / written at once, as I've mentioned. So basically, you read / write an 8 bit value from / to the port and it reads / sets all the 8 corresponding pins. E.g. Arduino has PORTB and PORTD with digital I/O pins and it allows you to access them both as individual pins like what you propose here and as the ports as a whole, 8 bits at once.

An example of a device that you want to control this way is e.g. a LCD character display like the one that you can see here: https://www.arduino.cc/en/Tutorial/HelloWorld, LED 7 segment displays, good old parallel port devices, etc.

jkotas: The APIs proposed here cover the beginner scenarios. They do not cover much beyond that.

Take a look at what kind of functionality the existing RPi libraries expose. http://abyz.me.uk/rpi/pigpio/python.html is the one I have been using recently. @janvorli is talking about `read_bank` / `write_bank` APIs from the intermediate set.

2018/System.Device.Gpio/README.md
terrajobst: I might have written this up incorrectly. AFAIK the suggestion was to expose concrete types to allow for more specialized APIs (as opposed to force the same API surface on everyone). I think you answered it by "that's what the driver is for".

In general, don't take these open questions as me or @dotnet/fxdc telling you what to do; whenever I note questions it is genuinely meant as merely "consider this feedback but you're the domain expert, so we yield to your expertise". The only ask we have is that you have a good explanation beyond "*shrug* didn't like it" ;-)

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: I see, yeah I believe that at least for now that type of functionality is out of scope, but it definitely sounds interesting so that we could add it later on.

terrajobst: @janvorli I've added a note about this requirement.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+### GpioDriver classes
+
+``` C#
+public abstract class GpioDriver
+{
+    protected GpioPin OpenPin(int pinNumber, PinMode mode);
+    protected bool TryOpenPin(int pinNumber, PinMode mode, out GpioPin);
+    protected void ClosePin(int pinNumber);
+    protected int Read(byte[] buffer, int offset, int count);
+    protected void Write(byte[] value, int offset, int count);
```
stephentoub: Why do these take `byte[]` rather than `Span<byte>`?  Seems like it's an advanced scenario to implement a custom driver, and we should be ok with such developers needing to work with spans instead of byte[]s.  Then the higher level APIs could optionally pass in either a byte[] or a span or some such thing, because the driver could support any of them, but if these take byte[], we'll be limited in what exposed APIs can be supported.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+IDisposable { public IDictionary<int, PinMode> Pins { get; } public void
+Write(byte[] value, int offset, int count); public int Read(byte[] buffer, int
+offset, int count); public void Dispose(); }
+
+### GpioDriver classes
+
+``` C#
+public abstract class GpioDriver
+{
+    protected GpioPin OpenPin(int pinNumber, PinMode mode);
```
stephentoub: Are all of these supposed to be abstract?  I'm not clear on how this is extended if they're not abstract.

2018/System.Device.Gpio/System.Device.Gpio.md
```diff
@@ -0,0 +1,100 @@
...
+    public int PWMRange { get; set; }
+    public event EventHandler<EventArgs> ValueChanged;
+    public int AnalogRead();
+    public int AnalogReadWait(TimeSpan timeout);
+    public void AnalogWrite(int value);
+    public new void Dispose();
+    public bool ReadValue();
+    public bool WaitForEvent(TimeSpan timeout);
+    public void WriteValue(bool value);
+  }
```
stephentoub: Similarly, if I'm implementing a custom driver, I don't need to be able to override functionality in GpioPin as well?  Wondering whether any of these should be virtual.

2018/System.Device.Gpio/System.Device.Gpio.md
```diff
@@ -0,0 +1,100 @@
...
+    public static void DefaultDriver_TurnLEDOn();
+    public static void DoSomething(object sender, EventArgs args);
+    public static void InitializeLightTurnedOn();
+    public static void PullUpButton();
+    public static void RaspberryPiDriver_TurnLEDOn();
+    public static void ReadButton();
+    public static void RegisterACallbackWhenButtonPressed();
+    public static void WaitForAButtonToBePressed();
+  }
+}
```
stephentoub: This samples namespace isn't actually shipping, right?

2018/System.Device.Gpio/System.Device.Gpio.md
```diff
@@ -0,0 +1,100 @@
...
+    Unknown = 8,
+  }
+  public enum PinNumberingScheme {
+    Board = 0,
+    Logic = 1,
+  }
+  public enum PWMMode {
+    BALANCED = 1,
+    MARK_SPACE = 0,
+  }
```
stephentoub: What's up with the all-caps in PWMMode's values?  And should it be PwmMode?  Or something less acronymy altogether?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+```C#
+using (GpioController controller = GpioController.Open())
+{
+    GpioPin button = new GpioPin(controller, 13, PinMode.Input);
+    GpioPin led = new GpioPin(controller, 18, PinMode.Output);
+    while (true)
+    {
+        PinValue buttonValue = button.ReadValue();
+        led.WriteValue(buttonValue);
+    }
```
stephentoub: Is this effectively a spin loop, such that we repeatedly read the current state and write the value without any blocking?  Or does ReadValue wait for something to change?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+
+## Proposed API
+
+Following are the proposed APIs.
+
+### GpioController class
+``` C#
+public class GpioController : IDisposable
+{
+    protected GpioDriver driver { get; }
```
stephentoub: driver => Driver

terrajobst:
@stephentoub 

Our API review didn't go into that much detail yet because we were still reviewing the fundamental architecture. I suggest we let @joperezr come back with an updated proposal where we go over every single API.

stephentoub:
Yup, thanks, I was just leaving notes about things I didn't get to comment on in the review today. :)

2018/System.Device.Gpio/System.Device.Gpio.md
terrajobst: Good point. Removed from API Ref.

terrajobst:
Cool, so I guess no objections with merging this as-is for now?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: It doesn't wait, this is just a sample that mimics what python uses for hello world scripts. In reality, you would probably want to either subscribe for a change event callback or you would at least set a debounce timeout.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: yes, They are all supposed to be abstract. This used to be an interface instead, but we got feedback suggesting to move it into an abstract class, so I changed the type but didn't go and change the members.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: For these guys, we were trying to have similar functionality to the basic members of FileStream. That said, I agree with you that `Span<byte>` would make more sense to give more flexibility to the developers of new drivers.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    // Driver that implements the GpioDriver actions by using SYSFS
+}
+
+public class RaspberryPi3Driver : GpioDriver
+{
+    // Driver that implements the GpioDriver actions by using the Raspberry Pi's
+    // BCM2835 Peripheral registers.
+}
+
+public class WindowsIOTDriver : GpioDriver
```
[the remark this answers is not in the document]
joperezr: I'm not 100% sure about that. The way we designed this, you could think of a case where there is a new type of sensor you want to plug in to your raspberry pi 3, and you want to have built-in support on the driver to read/write to the sensor or do more complex operations. In that case, I would imagine people writing a `MySensor` class that implements `GpioConnection` and a driver `RaspberryPi3WithMySensorSupportDriver` that derives from `RaspberryPi3Driver`. We are not yet sure on how common will that be, but we thought it might be common for sensor developers to provide their own Sensor class and driver that extends our drivers.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
jkotas: Also, it would be useful to have a real example of a driver that one creates using this extensibility point. I have no idea how this is meant to be used.

2018/System.Device.Gpio/System.Device.Gpio.md
```diff
@@ -0,0 +1,100 @@
...
+    public static GpioController Open(GpioDriver driver, PinNumberingScheme numbering=(PinNumberingScheme)(1));
+    public static GpioController Open(GpioDriverKind kind, PinNumberingScheme numbering=(PinNumberingScheme)(1));
+    public static GpioController Open(PinNumberingScheme numbering=(PinNumberingScheme)(1));
+  }
+  public abstract class GpioDriver {
+    protected GpioDriver();
+    protected internal void ClosePin(int pinNumber);
+    protected internal GpioPin OpenPin(int pinNumber, PinMode mode);
+    protected internal int Read(byte[] buffer, int offset, int count);
+    protected internal bool TryOpenPin(int pinNumber, PinMode mode);
```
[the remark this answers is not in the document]
joperezr: Yes, the only difference is that one throws and the other doesn't. The real reason why we wanted to have the OpenPin method as well, was to match the surface area that other frameworks have like the one for Python. That way developers would be able to port their python scripts to .NET easily. That said, I think it is trivial for them to just change their code to use `TryOpenPin` instead, so I like your idea of just removing `OpenPin`.

2018/System.Device.Gpio/System.Device.Gpio.md
```diff
@@ -0,0 +1,100 @@
...
+  public enum GpioDriverKind {
+    HummingboardEdge = 4,
+    ODroid = 3,
+    RaspberryPi3 = 0,
+    Unix = 1,
+    WindowsIOT = 2,
+  }
+  public class GpioPin : GpioConnection {
+    public GpioPin(GpioController controller, int gpioNumber, PinMode mode);
+    public GpioPin(GpioController controller, int gpioNumber, PinMode mode, bool initialValue);
```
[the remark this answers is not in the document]
joperezr: well, on the one where we don't set an initial value, we don't really have default, we just don't touch the bit that controls the value. That means that if somebody initialized a pin (connected to a led) left the pin on, and then disposed the pin, he could later go and initialize the same pin without an initial value, and the pin would still be on. The constructor with the initialValue would actually go in and flip the bit to whatever you decided to have on/off.

2018/System.Device.Gpio/System.Device.Gpio.md
```diff
@@ -0,0 +1,88 @@
...
+    public bool WaitForEvent(TimeSpan timeout);
+    public void WriteValue(bool value);
+  }
+  public enum PinEvent {
+    FallingEdge = 1,
+    RisingEdge = 2,
+  }
+  public enum PinMode {
+    I2C = 6,
+    Input = 3,
```
janvorli: It seems that based on the other enums namings, it should be PullDown and PullUp (without the underscore).
One more thing I've just noticed. The Pull_Down and Pull_Up are supposed to mean "input with pull down / pull up", right? Maybe we should name them InputPullDown and InputPullUp then to make it obvious.

2018/System.Device.Gpio/System.Device.Gpio.md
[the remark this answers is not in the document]
jkotas: > was to match the surface area that other frameworks have like the one for Python.

It would be useful to mention the Python frameworks that you are trying to match.

I have used two different ones:
- `RPi.GPIO`: It is the first one in the search results for https://www.bing.com/search?q=raspberry+pi+gpio+python
- `pigpio`: It is good for doing more advanced stuff.

Neither of them have objects representing pins. They just have a simple flat surface where you refer to pin using numbers, like `GPIO.input(23)`. Having used the simple flat surface, the objects representing pins feel both over-engineered and not powerful enough at the same time. It may be just that these APIs are not for me.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Blinking LED
+
+``` C#
+using (GpioController controller = GpioController.Open())
+{
+	GpioPin pin = new GpioPin(controller, 18, PinMode.Output);
+	while (true)
+	{
+	    pin.WriteValue(PinValue.High);
```
KrzysztofCwalina: I have to say "WriteValue" seems strange in this context. Would Set(PinValue.High) be better? If we want to stay with Write, can we drop the Value suffix? i.e. pin.Write(PinValue.High).

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+### Waiting for a button to be pressed
+
+```C#
+using (GpioController controller = GpioController.Open())
+{
+    GpioPin button = new GpioPin(controller, 13, PinMode.Input);
+    button.EnableRaisingEvents = true;
+    button.NotifyFilter = PinEvent.RisingEdge;
```
KrzysztofCwalina: What's the default for NotifyFilter? In particular, will I get any events if I don't set this property?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+GpioController controller = GpioController.Open(); 
+```
+
+### Using a different pin numbering scheme
+
+We will support two different types of numbering schemes: Board and Logical. The
+Board numbering is the physical number of where the pin is located, and the
+Logical is where the pin is represented by the Gpio number.
+
+``` C#
```
KrzysztofCwalina: The following 3 lines of code don't seem to be updated based on the discussion yesterday.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+## Proposed API
+
+Following are the proposed APIs.
+
+### GpioController class
+``` C#
+public class GpioController : IDisposable
+{
+    protected GpioDriver driver { get; }
+    public static GpioController Open(PinNumberingScheme numbering = PinNumberingScheme.Logic);
```
KrzysztofCwalina: I think we should consider removing the overload above. It messes up intellisense (as the position of the scheme parameter changes between overloads), yet the additional value is small IMO.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public static GpioController Open(GpioDriver driver, PinNumberingScheme numbering = PinNumberingScheme.Logic);
```
KrzysztofCwalina: I think the "numbering" parameter should be renamed. The parameter specifies a scheme, not numbering.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public static GpioController Open(GpioDriverKind kind, PinNumberingScheme numbering = PinNumberingScheme.Logic);
+    public PinNumberingScheme Numbering { get; set; }
```
KrzysztofCwalina: Same: it's not numbering, but rather a scheme, numbering scheme that is.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public IEnumerable<IGpioConnection> OpenConnections { get; }
```
KrzysztofCwalina: I think we should get rid of the name "connection". It adds to the concept count for no reason. There is not a single method called "connect", yet there are connections.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public int PinCount { get; }
+    public bool isPinOpen(int pinNumber);
```
KrzysztofCwalina: Casing

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public bool isPinModeSupported(int pinNumber, PinMode mode);
+    public void Dispose();
+    public void CloseAllPins();
```
KrzysztofCwalina: I thought we agreed to just use Dispose.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+}
+
+public enum GpioDriverKind
+{
+    RaspberryPi3,
+    Unix,
+    WindowsIOT,
```
KrzysztofCwalina: IOT should be cased "Iot", but that looks strange, so maybe we call it WindowsDevice?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public bool isPinOpen(int pinNumber);
+    public bool isPinModeSupported(int pinNumber, PinMode mode);
+    public void Dispose();
+    public void CloseAllPins();
+}
+
+public enum GpioDriverKind
+{
+    RaspberryPi3,
+    Unix,
```
KrzysztofCwalina: Why WindowsIOT but just Unix? Why not UnixIOT?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public PinNumberingScheme Numbering { get; set; }
+    public IEnumerable<IGpioConnection> OpenConnections { get; }
+    public int PinCount { get; }
+    public bool isPinOpen(int pinNumber);
+    public bool isPinModeSupported(int pinNumber, PinMode mode);
+    public void Dispose();
+    public void CloseAllPins();
+}
+
+public enum GpioDriverKind
```
KrzysztofCwalina: Should we call it XxxGpioDriver? e.g. StandardGpioDriver, BuiltInGpioDriver, etc.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+{
+    RaspberryPi3,
+    Unix,
+    WindowsIOT,
+    ODroid,
+    HummingboardEdge
+}
```
KrzysztofCwalina: Do we need a query method to determine if a driver is supported? e.g.
if(GpioController.IsSupported(GpioDriverKind.Unix)) { ... }

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+### GpioPin class
+``` C#
+public class GpioPin : GpioConnection
+{
+    // Core Functionality of a Pin
+    public GpioPin(GpioController controller, int gpioNumber, PinMode mode);
```
KrzysztofCwalina: We seem to be calling a single concept pinNuber and gpioNumber. If that's indeed a case, we should pick one.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public GpioPin(GpioController controller, int gpioNumber, PinMode mode, bool initialValue);
+    public new void Dispose();
+    public bool ReadValue();
```
KrzysztofCwalina: Why in't this just a get/set Value property?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public void WriteValue(bool value);
+    
+    // PWM
+    public int PWMFrequency { get; set; }
```
KrzysztofCwalina: Casing

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
shaggygi: What about WinRt (assuming it would be using those specific APIs)?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public float PWMDutyCycle { get; set; }
+    public int PWMRange { get; set; }
+    public PWMMode PWMMode { get; set; }
+
+    // Analog RW
+    public int AnalogRead();
+    public void AnalogWrite(int value);
+    public int AnalogReadWait(TimeSpan timeout);
```
KrzysztofCwalina: Can we support cancellation?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public void WriteValue(bool value);
+    
+    // PWM
+    public int PWMFrequency { get; set; }
+    public float PWMDutyCycle { get; set; }
+    public int PWMRange { get; set; }
+    public PWMMode PWMMode { get; set; }
+
+    // Analog RW
+    public int AnalogRead();
```
KrzysztofCwalina: AnalogValue property?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public void AnalogWrite(int value);
+    public int AnalogReadWait(TimeSpan timeout);
+
+    // Eventing
+    public TimeSpan DebounceTimeout { get; set; }
+    public bool EnableRaisingEvents { get; set; }
+    public PinEvent NotifyFilter { get; set; }
```
KrzysztofCwalina: How many values are there in NotifyFilter? If not too many, should we just have separate events for each?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public event EventHandler<EventArgs> ValueChanged;
+    public bool WaitForEvent(TimeSpan timeout);
```
KrzysztofCwalina: Can we support cancellation?

2018/System.Device.Gpio/System.Device.Gpio.md
```diff
@@ -0,0 +1,100 @@
+# System.Devices.Gpio
+
+```C#
+namespace System.Device.Gpio {
+  // Immo Landwerth: Why is this needed? Could we expose this later?
+  public abstract class GpioConnection : IDisposable {
+    protected GpioConnection();
+    public IDictionary<int, PinMode> Pins { get; }
+    public void Dispose();
+    public int Read(byte[] buffer, int offset, int count);
```
KrzysztofCwalina: Should we use Span here?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
```diff
@@ -0,0 +1,273 @@
...
+    public int PinCount { get; }
+    public bool isPinOpen(int pinNumber);
+    public bool isPinModeSupported(int pinNumber, PinMode mode);
+    public void Dispose();
+    public void CloseAllPins();
+}
+
+public enum GpioDriverKind
+{
+    RaspberryPi3,
```
janvorli: I actually wonder why this is an enum. There are many different devices, not just the few listed here. Should we rather support pluggable drivers and use just string as GpioDriverKind so that new drivers can be easily added both by us and by third parties without a need to modify this part of the API?

2018/System.Device.Gpio/System.Device.Gpio.md
```diff
@@ -0,0 +1,100 @@
+# System.Devices.Gpio
+
+```C#
+namespace System.Device.Gpio {
+  // Immo Landwerth: Why is this needed? Could we expose this later?
+  public abstract class GpioConnection : IDisposable {
+    protected GpioConnection();
+    public IDictionary<int, PinMode> Pins { get; }
```
KrzysztofCwalina: I still find this confusing. A single pin inherits from this type. The pin knows its number (the int), and so it seems to me there should simply be a property there returns PinMode  of the current pin. Also, if we do want to support reading pinmode of an arbitrary pin, could this be a method GetPinMode(int pinNumber)? This way we don't have to allocate a dictionary.

2018/System.Device.Gpio/System.Device.Gpio.md
```diff
@@ -0,0 +1,100 @@
...
+    Unix = 1,
+    WindowsIOT = 2,
+  }
+  public class GpioPin : GpioConnection {
+    public GpioPin(GpioController controller, int gpioNumber, PinMode mode);
+    public GpioPin(GpioController controller, int gpioNumber, PinMode mode, bool initialValue);
+    public TimeSpan DebounceTimeout { get; set; }
+    public bool EnableRaisingEvents { get; set; }
+    public PinEvent NotifyFilter { get; set; }
+    public float PWMDutyCycle { get; set; }
```
KrzysztofCwalina: Casing.

2018/System.Device.Gpio/System.Device.Gpio.md
```diff
@@ -0,0 +1,100 @@
...
+  }
+  public enum PinEvent {
+    FallingEdge = 1,
+    RisingEdge = 2,
+  }
+  public enum PinMode {
+    I2C = 6,
+    Input = 3,
+    Output = 4,
+    Pull_Down = 0,
```
KrzysztofCwalina: We don't use underscores in APIs.

2018/System.Device.Gpio/System.Device.Gpio.md
MarkPflug: Should GpioPin be sealed then?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: I think I would like to stick with Read/Write if possible since I think they are easier to find than Set and Clear. I'll remove the Value suffix.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: No, Default is `0x00` so you wouldn't be subscribed for any notifications.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: Nothing here has been updated, these are the docs as were presented yesterday, I'm currently working on addressing all of the comments and will push a new iteration.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: so would you instead not having it as a parameter in a factory method and instead a value that could be set later? Or would you rather have an extra factory method that gets this property set as well?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: what about `numberingScheme` instead?

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: this was kind of part of the Base class of GpioPin discussion, but given that we decided to get rid of that then this will change as well.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: yes. This isn't updated after our meeting yesterday.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: Yes, we actually had both options. So we had factory methods that would get you a controller based on this enum kind, or alternatively you could pass in your own driver class that would take care of the pin operations. We decided to have an enum since that would only contain the drivers that ship with the library, but custom drivers could be used as well by passing those to the factory method. In any case, we decided the enum provided little to no value, so we will get rid of this.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: mm I'm not sure I understand where would you use this. Do you mean that we would use it in order to have code that works for both linux os or windows like:
```c#
GpioController controller;
if (GpioController.IsSupported(GpioDriverKind.Unix))
{
    controller = GpioController.Open(GpioDriverKind.Unix);
}
else
///... initialize on some other way
```

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: originally the intention for this was that `pinNumber` would be the number based on the `PinNumberingScheme` of the controller, and `gpioPinNumber` would be the logical number always. I do agree however that this is confusing, so I'll clean it up.

2018/System.Device.Gpio/System.Device.Gpio.Proposal.md
joperezr: based on your other piece of feedback, this is now Write() and Read() which I think makes more sense.

mwpowellhtx:
Just following the design review conversation the past couple of postings. Interesting, but if I may, a couple of things:

* The whole ``AddCallback``/``RemoveCallback`` paradigm just feels a lot like *Java* (and if you did not cringe at that... :-1:). I wondered, why not just expose a true .NET C# ``event``, this was mentioned later as well in terms of ``event`` shape. I agree with that critique, use ``event``.

* But moreover, so much focus on a *single pin* is a mistake. It really is. I am speaking from experience throughout my career working with GPIO, serially over RS232, embedded using I2C. I worked through C/C++ using ``bitset`` to engage with the IO. :+1: I do not recall exactly if that was *Boost DynamicBitset* or *std::bitset*, but you get the idea.

Really, *"Blinking LED's"*, while interesting, is just one application, and a rather academic one at that. I did that, I also scanned for colors, I turned a light source on/off, I varied the light source intensity, I also engaged with a custom "oven" of sorts, literally with a BTU register and closed feedback loop.

In other words, in practical terms, you really do not work with a single pin at a time, ever, even for the *"blinking LED"* exercise, never mind devoting a single file to a single pin. That just doesn't happen in real world applications. Instead you want to work with the entire mask, whether at the kernel level or user level, do the twiddling, ``event``, ``async``, etc, update the register if it was an output request, and so on, and move on.

What I did do usually was devote a module to a given I2C controller, for instance. Then if there were elements of that which required register level mapping, that was my API for that instance, and I would customize the bitset/register scaffold accordingly.

joperezr:
For the Add/Remove callback there is actually some precedence in the framework for this as well: https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadpool.registerwaitforsingleobject?view=netcore-2.1
I agree it doesn't look a lot like C#, but we just feel that it will be more discoverable this way which is why we went with that idea to start with. That said, this is very good feedback and we can always move to a different model.

I agree that most apps won't be dealing with GPIO directly, but instead talking to most meaningful protocols like SPI and I2C (which are also part of this proposal). That said, there are some devices (like the LCM1602C display for example) that do deal with GPIO directly as they don't use more advance protocols. For things like this one, I do expect people to use GPIO pins directly. This initial proposal is to provide the base implementation of communication between these ports, but eventually, this can grow into something in a higher layer that will deal with whole masks instead.

mwpowellhtx:
@joperezr ...

> we just feel that it will be more discoverable

*Discoverable* how? By a web search of the MSDN documentation? By *IntelliSense*? FWIW, one professional to another, events are no less discoverable than methods, properties, etc.

The only precendent I would draw from *Threading* is the potential want to operate GPIO in an ``async`` manner, which I think makes perfect sense. That's about all. I wouldn't stand on that assumption as though it were gospel, however.

Good point re: *LCM1602C*. Good point.

Anyway, that's my two cents speaking from experience.

joperezr:
> Anyway, that's my two cents speaking from experience.

First of all, we really appreciate the feedback so keep that coming 😄 

> Discoverable how? 

Given that we have only a controller class, and not a pin object class, and that we want to be able to register event handlers for individual pins, the option of doing this with eventing would be to expose some sort of indexed property which used events, like:
```c#
    controller.PinEvents[pinNumber, evenType].Raising += MyHandler;
    controller.PinEvents[pinNumber, evenType].Falling += MyHandler;
```
What I mean about more discoverable is that it might not be super obvious (through intellisense) that this is the way to add or register a callback for an event, where if instead I just type "controller." and see that there is a method called RegisterCallbackForPinValueChangedEvent I will immediately know what that does. As I said, we are still totally open using the pattern above, which is why we appreciate the feedback saying that it is not any less discoverable than the current approach.

mwpowellhtx:
@joperezr It's an argument as much for well-named, in this case, the suggestion is C# ``event``, as it is for, a robust enough *Xml* documentation, more than *"this is the event"*, that seems to constitute so much of the MSDN documentation, for instance.

PR #69, opened 2018-10-18
SequenceReader<T>: Round 2

terrajobst:


2018/System.Buffers.SequenceReader/README.md
```diff
@@ -49,3 +51,46 @@ Status: **Needs more work** |
...
+## Session 2
+
+### SequenceReader\<T>
+
+* Should we remove the `unamanged` constrained and replace with a `.ctor` check?
+    - Done for pinning and being able to take a pointer?
+* Can we have some sample code that uses a `ReadOnlySpan<T>`-based reader vs. a
+  `ReadOnlySequence<T>`-based reader?
+    - We can then compare the per gains
+    - We can then compare the usability gains
```
stephentoub: And specifically (since I'm not sure I was clear in the meeting), here's what I was hoping to see.  Given some common scenario that reads from / parses data from a pipeline:
- Use the BufferReader that only supports span, and have a single method (written as we would in production code as best we can) with all of the relevant boilerplate that handles interaction both with the pipeline and with managing sequences/calls to the span-based reader.
- Use the BufferReader that supports sequence, and have a single method (written as we would in production code as best we can) with all of the relevant boilerplate that handles interaction both with the pipeline and with the sequence-based reader.

I'd like to us to be able to look at the code the provides the same functionality and the benchmarks on it and compare side-by-side both usability and performance.  If we could do the test as part of kestrel (e.g. two different runnable commits off of master) and look at plaintext benchmarks and what the kestrel code looks like both ways, even better.

If the version with the span-based reader is totally incomprehensible and provides only modest gains, then that leads us one way. If the version with the span-based reader is only mildly more cumbersome and provides significant gains, then that leads us another way. Etc. I just want to make sure we're all in agreement on the goals, seeing the same data, and then coming to a joint conclusion.

2018/System.Buffers.SequenceReader/README.md
```diff
@@ -49,3 +51,46 @@ Status: **Needs more work** |
...
+* `AtLastSegment`. The logic outlined in the comment is wrong, you need to honor
+  position, and not check whether there is next segment.
+* `Allocator` seems odd. If we need it, should we promote the delegate or move
+  it elsewhere? We can't use a `Func` due to the generic instantiation issue
+  with `Span<T>`.
```
stephentoub: We're starting to see a need for more and more specialized delegates, each time we have a need for a new combination of span arguments and return types.  Should we be looking at ways of actually allowing span to be used as generics in constrained cases?  Is there anything we could do there meaningfully in the 3.0 timeframe?

2018/System.Buffers.SequenceReader/README.md
```diff
@@ -49,3 +51,46 @@ Status: **Needs more work** |
...
+  latter isn't clear whether it's relative to the current position vs. the
+  buffer's start.
+* We may want to make `advancePastDelimiter` different methods to avoid the perf
+  issue due to the extra branches.
+* The struct has to be passed by `ref` or you have bugs. Also, since it's large,
+  passing it by value would also tank perf.
+
+### SequenceReaderExtensions
+
+* All methods have a limit of reading 128 bytes maximum
```
stephentoub: 128 bytes or 128 Ts?  Presumably the latter.

2018/System.Buffers.SequenceReader/README.md
```diff
@@ -49,3 +51,46 @@ Status: **Needs more work** |
...
+  `ReadOnlySequence<T>`-based reader?
+    - We can then compare the per gains
+    - We can then compare the usability gains
+* `AtLastSegment`. The logic outlined in the comment is wrong, you need to honor
+  position, and not check whether there is next segment.
+* `Allocator` seems odd. If we need it, should we promote the delegate or move
+  it elsewhere? We can't use a `Func` due to the generic instantiation issue
+  with `Span<T>`.
+    - Should the allocator be passed in as argument? Storing it as a field might
+      cause additional memory barrier
```
benaadams: Also increases size when returned by value (all returns) and when copied by value (for save/rewind points)

2018/System.Buffers.SequenceReader/README.md
```diff
@@ -49,3 +51,46 @@ Status: **Needs more work** |
...
+    - This allows stack allocated buffers to be used and seemed rationale
+    - Might be too small for specialized scenarios, such as many leading zeros
+    - We cannot really expose APIs with sizes as this would likely destroy perf
+* No APIs can produce strings, we should probably add that as dealing with
+  things going across buffers is hard and we don't want folks to go through an
+  intermediary array.
+
+## BufferExtensions
+
+* Can't we make this an instance method?
```
benaadams:
```csharp
// returns false if data is more than one span
public bool TryAsSpan<T>(out ReadOnlySpan<T> span);
// copyBuffer should be >= .Length
public ReadOnlySpan<T> AsSpan<T>(in this ReadOnlySequence<T> source, Span<T> copyBuffer);
```

2018/System.Buffers.SequenceReader/README.md
JeremyKuhne: T's for the byte/char extensions- this only applies to the *Parse extensions.

2018/System.Buffers.SequenceReader/README.md
benaadams: Also should there be a return flag to indicate if the `copyBuffer` was used? 

Otherwise there is ambiguity. Is there is a writable version (or copy) in the copyBuffer? Can the copyBuffer be reused prior to consuming the `ReadOnlySpan` (i.e. is it part of the ReadOnlySpan or was it unused)

benaadams: To resolve the ambiguity; especially around exceptions, maybe it should be:
```csharp
// returns false if data is more than one span
public bool TryAsSpan<T>(out ReadOnlySpan<T> span);
// destination should be >= .Length
public void CopyTo(Span<T> destination);
```
Then the pattern would be something like:
```csharp
T[] scratch = null;

if (!seq.TryAsSpan(out ReadOnlySpan<T> roSpan)
{
    scratch = ArrayPool<T>.Shared.Rent(seq.Length)
    seq.CopyTo(scratch);
    roSpan = scratch;
}

// Do something with roSpan

if (scratch != null) ArrayPool<T>.Shared.Return(scratch);
```

stephentoub: > Also should there be a return flag to indicate if the copyBuffer was used? Otherwise there is ambiguity

It seems really rare that you'd care, though, right?  Presumably you'd just assume you can't use the supplied scratch buffer again until you're done with the returned span.  And in the rare case where you actually cared, you could use Span.Overlaps to determine whether the returned span is  part of the scratch buffer or not.

That said, I agree in general that this pattern is confusing.

benaadams: Problem is if you want to modify the returned data (mask/unmask etc) then with
```csharp
public ReadOnlySpan<T> AsSpan<T>(Span<T> copyBuffer);
```
You need one scratch buffer for the call (which may or may not be used); then a second scratch buffer to copy the returned span into; whereas it would be easier just to jump straight to `.CopyTo`

For just reading; then the scratch buffer is unnecessary for the fast-path; but you need a `.Length` sized buffer allocated just in case and if the whole `ReadOnlySequence` is being read then the scratch buffer is unlikely to be used for more than a single read (e.g. returned to pool prior to awaiting for more data).

Also question on what happens if `.Length` is greater than 2Bn elements.

stephentoub: As I said, I agree it'd be better not to force the caller to allocate until they know they have to. I'm not arguing against that.  I was replying to your comments about knowing whether the buffers are the same so that you would know whether you could reuse the scratch buffer while still using the returned span, and I was saying I think you'd just always assume you couldn't use the scratch buffer until you were done with the returned span.  That's all.

benaadams: I'm just thinking aloud :)

benaadams: I'm just thinking aloud :)

To cope with more than 2Bn elements should `.CopyTo` then be something like:
```csharp
public void CopyTo(Span<T> destination, out isComplete);
// or
public bool CopyTo(Span<T> destination);
```
Which would then also imply that the span could be shorter than `.Length` (2Bn aside); so not throw an exception, and if it returns `false`, to get the next the `SequenceReader` should be moved on by `.Advance(destination.Length)`?

JeremyKuhne: There is value in having a single method option for this if you're perf critical. I believe taking an optional `Allocator` makes this possible while still giving full context as to what happened (I think @davidfowl suggested as much?).

``` C#
// What we currently have shipped:
public static T[] ToArray<T>(in this ReadOnlySequence<T> sequence);
public static void CopyTo<T>(in this ReadOnlySequence<T> source, Span<T> destination);

// ** One set of options **

// Slices if possible, otherwise copies to new[] 
ReadOnlySpan<T> SliceOrCopy();
// Or maybe?? (meh)
ReadOnlySpan<T> SliceOrToArray();
// Slices if possible, otherwise copies to Allocator result
ReadOnlySpan<T> SliceOrCopyTo(Allocator<T> allocator);
// Or possibly just? (if null, does new T[])
// (Would want to measure if the `allocator == null ? new T[] : allocator()` condition has much of an impact)
ReadOnlySpan<T> SliceOrCopyTo(Allocator<T> allocator = null);
```

Note that I don't think @benaadams suggestions are mutually exclusive to these.

@pakrym/@davidfowl do you have any feedback on this discussion?

benaadams: I think the allocator param is valid; however it might need further thinking though. Would it need an associated data structure to track allocations (outside of vanilla `new T[]`)? 

For example, you couldn't just allocate from `ArrayPool<T>.Shared` as you wouldn't be tracking the arrays to return; so would perhaps need an additional object? Would something with a `using` block work? Would want to avoid a delegate allocation per call etc.

JeremyKuhne: > so would perhaps need an additional object? 

I would think that we could write something like a disposable ArrayPoolTracker<T> that would allow you to manage the lifetime of rented arrays through an Allocator.

> Would want to avoid a delegate allocation per call etc.

Yeah, the trickiness of avoiding delegate allocations concerns me. I wonder if an abstract class would be better than a delegate- haven't thought that one through.  Might be a totally stupid idea.

> can you share numbers on the perf here for "if you're perf critical"?

Yes, I'm working through a whole bunch of experiments and will post as soon as I can.

JeremyKuhne: > And if ReadOnlySequence<T> has multiple segments, you're invariably going to need to copy, unless you got super luck

Highly unlikely that you will cross segments in the parsing scenarios I've been looking at (Kestrel). It will almost always be a single segment that would be returned from `SequenceReader`.

I'm still trying to characterize all of the variables going into the performance of the above. In the simplest, copy to new array, single segment case I'm seeing an overhead of 18% (over 60% if I don't inline). 

Here are the extension methods I'm testing with:

``` C#
// Inlining this one makes it worse
public static ReadOnlySpan<T> SliceOrCopy<T>(in this ReadOnlySequence<T> sequence)
{
    return sequence.IsSingleSegment ? sequence.First.Span : sequence.ToArray();
}

[MethodImpl(MethodImplOptions.AggressiveInlining)]
public static bool TryAsSpan<T>(in this ReadOnlySequence<T> sequence, out ReadOnlySpan<T> span)
{
    bool single = sequence.IsSingleSegment;
    span = single ? sequence.First.Span : default;
    return single;
}
```

And the tests:

``` C#
[Benchmark]
public ReadOnlySpan<byte> SliceOrCopy_Array()
{
    return _arraySequence.SliceOrCopy();
}

[Benchmark]
public ReadOnlySpan<byte> TwoStage_Array()
{
    if (!_arraySequence.TryAsSpan(out ReadOnlySpan<byte> span))
    {
        byte[] scratch = new byte[(int)_arraySequence.Length];
        _arraySequence.CopyTo(scratch);
        span = scratch;
    }

    return span;
}
```

The sequence is `new ReadOnlySequence<byte>(new byte[10_000])`.

I'm looking into the "Allocator" pattern now...

JeremyKuhne: Ugh, copied the wrong code- updating...

JeremyKuhne: Here is using `Allocator` to new up an array if needed:

``` C#
private Allocator<byte> _allocator = (length) => new byte[length];

[Benchmark]
public ReadOnlySpan<byte> SliceOrCopy_AllocatorArray()
{
    return _arraySequence.SliceOrCopy(_allocator);
}

[Benchmark]
public ReadOnlySpan<byte> TwoStage_Array()
{
    if (!_arraySequence.TryAsSpan(out ReadOnlySpan<byte> span))
    {
        Span<byte> scratch = new byte[checked((int)_arraySequence.Length)];
        _arraySequence.CopyTo(scratch);
        span = scratch;
    }

    return span;
}

public static ReadOnlySpan<T> SliceOrCopy<T>(in this ReadOnlySequence<T> sequence, Allocator<T> allocator)
{
    if (sequence.IsSingleSegment)
    {
        return sequence.First.Span;
    }

    Span<T> span = allocator(checked((int)sequence.Length));
    sequence.CopyTo(span);
    return span;
}
```

```
                     Method |     Mean |     Error |    StdDev |   Median |
--------------------------- |---------:|----------:|----------:|---------:|
          SliceOrCopy_Array | 14.67 ns | 0.3196 ns | 0.5512 ns | 14.90 ns |
 SliceOrCopy_AllocatorArray | 15.75 ns | 0.3365 ns | 0.7596 ns | 16.22 ns |
             TwoStage_Array | 17.87 ns | 0.3825 ns | 0.6994 ns | 18.21 ns |
```

stephentoub: Thanks for running the numbers!

Seems like:
```C#
public static bool TryAsSpan<T>(in this ReadOnlySequence<T> sequence, out ReadOnlySpan<T> span)
{
    bool single = sequence.IsSingleSegment;
    span = single ? sequence.First.Span : default;
    return single;
}
```
can be branchless as:
```C#
public static bool TryAsSpan<T>(in this ReadOnlySequence<T> sequence, out ReadOnlySpan<T> span)
{
    span = sequence.First.Span;
    return sequence.IsSingleSegment;
}
```
since it doesn't matter what the output span is if the method returns false, and we're optimizing for the true case.  Does that make a difference?

Could also try reversing the signature, having it return a span and have the out be a bool, to avoid a write barrier,  but for that matter, why would we even need `SliceOrCopy`... the caller can do that themselves just as easily, and then decide what they want to do in the fallback case, rather than having to have methods that dictate it for them, no?

JeremyKuhne: Branchless made it worse:

``` 
BEFORE
           TwoStage_Array | 17.87 ns | 0.3825 ns | 0.6994 ns | 18.21 ns |  

AFTER
           TwoStage_Array | 18.73 ns | 0.3999 ns | 0.7894 ns | 19.18 ns |
```

Not terribly surprising- ripping the span out being costly is the only reason `SequenceReader` is ref after all. :) Flipping to return a span made it slightly better, but still slower than `TryAsSpan` and is pretty awkard imho:

```
           TwoStage_Array | 17.03 ns | 0.3595 ns | 0.5597 ns | 17.24 ns |
```

``` C#
            ReadOnlySpan<byte> span = _arraySequence.TryAsSpan(out bool success);
            if (!success)
            {
```

> why would we even need SliceOrCopy

I'm not sure what you're asking here? Having the `SliceOrCopy` methods period is due to the performance improvement in the normal single segment case. Having one without a delegate is again about performance- as crossing segments is so unlikely, `new T[]` is a good-enough solution for many. Given that it is expected to be common and is faster it makes sense to have an "even-faster" overload for "default" behavior.

stephentoub: Here are the results I get copying and pasting the benchmarks you shared:
```
            Method |     Mean |     Error |    StdDev |   Median |
------------------ |---------:|----------:|----------:|---------:|
 SliceOrCopy_Array | 18.25 ns | 0.3942 ns | 0.6695 ns | 18.04 ns |
    TwoStage_Array | 17.58 ns | 0.1277 ns | 0.1132 ns | 17.57 ns |
```
I ran it multiple times, the results always come out relatively the same, with the TwoStage_Array being faster than SliceOrCopy_Array.  Note that that's with:
```
private static ReadOnlySequence<byte> _arraySequence = new ReadOnlySequence<byte>(new byte[10_000]);
```
since it wasn't clear to me from what you wrote what _arraySequence was.  If instead change it to be an instance, the versions end up being almost the same perf-wise for me:
```
            Method |     Mean |     Error |    StdDev |
------------------ |---------:|----------:|----------:|
 SliceOrCopy_Array | 17.47 ns | 0.3707 ns | 0.3468 ns |
    TwoStage_Array | 17.31 ns | 0.3631 ns | 0.4036 ns |
```

We're talking nanosecond differences here, in either direction.  Maybe it's just minor differences on the machine.  Maybe it's the processor involved.  Maybe it's something else.  But let's be very careful not to add lots of APIs because we found a microbenchmark showed an improvement on one machine.

JeremyKuhne: Here is the complete code:

https://gist.github.com/JeremyKuhne/2eb803b80058fa8b108c81352bc31ba0

> I ran it multiple times, the results always come out relatively the same, 

Ditto:

```
                     Method |     Mean |     Error |    StdDev |   Median |
--------------------------- |---------:|----------:|----------:|---------:|
          SliceOrCopy_Array | 14.87 ns | 0.3224 ns | 0.7343 ns | 15.21 ns |
 SliceOrCopy_AllocatorArray | 16.30 ns | 0.3458 ns | 0.4247 ns | 16.41 ns |
             TwoStage_Array | 17.82 ns | 0.3756 ns | 0.5736 ns | 18.05 ns |
```

> But let's be very careful not to add lots of APIs because we found a microbenchmark showed an improvement on one machine.

Well, sure, let's not be blind. But lets also not throw this out because some machines don't get a boost. For the record, my run is on a 6 core i8700K with a 2.1 .NET Core console app and the current BenchmarkDotNet (CTRL-F5ed of course) on Windows.

> They're one-liners, using public APIs that anyone can use.

They are simple because these are meant to validate the overhead involved. The whole point of the Allocator is to allow plugging in _whatever_ you want- which may be significantly more complicated. The ArrayPool tracker I mentioned above is one such example. I plan to move onto 

The simple `SliceOrCopy_Array` with no arguments as it makes writing the typical case easy. It also helps discoverability of the fact that `.ToArray()` is potentially inefficient. I wouldn't toss it simply because someone can easily rewrite the code.

For the record, here is what I get plugging in your `OpenCoded`:

```
                     Method |     Mean |     Error |    StdDev |   Median |
--------------------------- |---------:|----------:|----------:|---------:|
                  OpenCoded | 14.52 ns | 0.3108 ns | 0.5018 ns | 14.72 ns |
          SliceOrCopy_Array | 14.95 ns | 0.3185 ns | 0.5904 ns | 15.21 ns |
```

Not as much as an improvement for me.

Having these methods does not preclude you from writing your own logic. Having them, however, allows easy piping of the Allocator approach (e.g. I can take one as input to `SequenceReader`), facilitates reusability of solutions, **and** makes things significantly faster on some machines. All of that makes it pretty compelling to me.

If you could try to run again with the code in the gist and share your configuration perhaps we might find out what it is that makes your run slower than mine.

stephentoub: > Having them, however

And is slower on some machines apparently, and clutters the surface area (which can actually make it harder for a consumer of the API to do the "right thing'"), and adds surface area that may encourage developers to allocate when they don't actually need to, and a developer who's really concerned about eeking out the best possible throughout arguably won't use them, anyway, and for everyone else a nanosecond difference doesn't matter, and they're barely shorter than just writing out the ternary manually. As for "piping the allocator", that could still be used by ToArray if it's desirable, so that doesn't seem to actually change anything.

I will try the shared code again this weekend on a few machines. But at the moment I'm not seeing the benefit here (for ToArray, yes, for SliceOrCopy, no). I would rather we teach developers about the few core APIs that are there and how to best use thm efficiently, saving additional APIs we add for when they're truly meaningful (like the general reader APIs you've been working on).

PR #70, opened 2018-10-19
GPIO, Round 2

terrajobst:


2018/System.Device.Gpio2/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Links:
+
+* https://gist.github.com/joperezr/5da1bde93e2985ecead2d453f9c1a34f
+* https://gist.github.com/joperezr/a293d30fc714b9e92756c10f1ea89afe
+
+## Notes
+
+* We'll want to expose other protocols too (a, b, c), hence we probably don't
+  want to name the assembly `System.Device.Gpip`. We need a more general name,
```
joperezr: NIT: `System.Device.Gpip` -> `System.Device.Gpio`

2018/System.Device.Gpio2/README.md
```diff
@@ -0,0 +1,34 @@
...
+    - This would also means we probably don't want to use the `System` namespace.
+    - For instance, `Microsoft.Device.Xxx`?
+* The `WaitForEvent` method is still prone to problem of lossy even as events
+  that occur between two calls of `WaitForEvent` will not be observed.
+    - Jose should explore what to here
+    - It would be good to see what WinRT did here (it seems they expose an
+      object that has a lifetime that manages a queue) and what Python does.
+    - Jan's proposal is set beginners on the path of `while (true) { /* check state */ }` pattern.
+    > Sure, the chip gets a bit hotter, but who cares?
+* We need to think about we can expose new pins, e.g. MCP23008, which is a GPIO
```
joperezr: Not sure what this sentence is supposed to mean?

2018/System.Device.Gpio2/System.Device.Gpio.Proposal2.md
```diff
@@ -0,0 +1,303 @@
...
+##### No pin object model
+
+[Link to a similar model in the framework](https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadpool.registerwaitforsingleobject?view=netcore-2.1)
+
+```C#
+using (GpioController controller = GpioController.GetController())
+{
+    int button = 5;
+    controller.OpenPin(button, PinMode.Input);
+    controller.AddCallbackForValueChangedEvent(button, PinEventTypes.Rising, OnPinValueChanged);
```
janvorli: Is there any device where you can get some kind of native callback when a pin changes? It seems that e.g. RPi only has a polling way to detect that an event has occurred on a pin (an example for one of the c libraries for GPIO can be seen here: http://www.airspayce.com/mikem/bcm2835/event_8c-example.html - there is a flag related to the pin that you clear and when the change you are interested in happens, the flag gets set until you clear it - so it is essentially similar to Windows events). If we wanted to support events like this, we would need to have some internal background thread checking for the events in a loop and then issuing the event. I wonder if it wouldn't be better to keep just the WaitForEvent API.

2018/System.Device.Gpio2/System.Device.Gpio2.md
```diff
@@ -0,0 +1,222 @@
...
+        // Eventing
+        public WaitForEventResult WaitForEvent(int pinNumber, PinEventTypes eventType, int timeout);
+        public async Task<WaitForEventResult> WaitForEventAsync(int pinNumber, PinEventTypes eventType, int timeout, CancellationToken cancellationToken);
+        public void AddCallbackForValueChangedEvent(int pinNumber, PinEventTypes eventType, PinChangeEventHandler callback);
+        public void RemoveCallbackForValueChangedEvent(int pinNumber, PinChangeEventHandler callback);
+        
+        // Misc functions
+        public bool isPinOpen(int pinNumber);
+        public bool isPinModeSupported(int pinNumber, PinMode mode);
+        public static int GetPinNumberFromConfig(string configurationName, int defaultValue);
```
janvorli: Can you please explain what this method is going to do? I am not sure where does the config come from.

2018/System.Device.Gpio2/System.Device.Gpio2.md
```diff
@@ -0,0 +1,222 @@
...
+        // Controller fields
+        public PinNumberingScheme NumberingScheme { get; }
+        public int PinCount { get; }
+
+        // Open and close operations on Pins
+        public void OpenPin(int pinNumber);
+        public void OpenPin(int pinNumber, PinMode mode);
+        public void OpenPin(int pinNumber, PinMode mode, PinValue value);
+        public void BatchOpenPins(IEnumerable<int> pinNumbers);
+        public void BatchOpenPins(IEnumerable<int> pinNumbers, PinMode mode);
```
janvorli: Why do we add two functions differing in just an additional parameter instead of the `mode` parameter having a default value and keeping just the two parameter version?

2018/System.Device.Gpio2/System.Device.Gpio.Proposal2.md
```diff
@@ -0,0 +1,303 @@
...
+
+#### Reading if a button is pressed
+
+##### No pin object model
+
+```C#
+using (GpioController controller = GpioController.GetController())
+{
+    int button = 4;
+    controller.OpenPin(button, PinMode.Input);
```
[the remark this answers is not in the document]
joperezr: We will also add the SetPinMode method, but OpenPin's functionality is not just to set its mode. There are some Internal Drivers that will require the actual pin to be exported for use (For example, the Unix driver which uses SYSFS). In this model, we need to first Open the pin, and then we can do things with it, like set its mode or perform read/write operations.

The ClosePin, would basically make sure it unexports the pin so that it goes into a clean state. This is very important, and if not done manually, it will be called on the controller.Dispose() for each opened pin.

2018/System.Device.Gpio2/System.Device.Gpio2.md
joperezr: This was discussed in the review and was supposed to help with the docker scenario. Basically the idea behind it is that if your app would run in a docker container(meaning that the actual program is probably already built in the image) then you would have a way to override the pins used for hooking up whatever device you would use. So for example, if one specific microcontroller where you wanted to run your app inside a container is already using one of the default pins, then you could override it with some configuration setup (this being either an environment variable or a config file.) In any case, during the Api Review it was decided that this sort of functionality shouldn't live with the GPIO API, and if a user needed this then they should write their own code which would take care of this.

2018/System.Device.Gpio2/System.Device.Gpio2.md
joperezr: This is because for some drivers (like the SYSFS driver) you can open a pin, but not set its mode. In this case, the pin will be opened using whichever mode it was last used for (or Input if it has never been used before) If we provide a default, then we would be overriding this behaviour.

PR #71, opened 2018-10-23
Xamarin.Essentials

terrajobst:


2018/Xamarin.Essentials/Xamarin.Essentials.md
```diff
@@ -0,0 +1,559 @@
...
+    public override int GetHashCode();
+    public static bool operator ==(BarometerData left, BarometerData right);
+    public static bool operator !=(BarometerData left, BarometerData right);
+  }
+  public static class Battery {
+    public static double ChargeLevel { get; }
+    // Immo Landwerth: Consider returning 1.0 when the device is using AC. This makes
+    //                 common case where folks check against a threshold much simpler.
+    //                 Consider percentage in the name as those are customary between
+    //                 0.0 and 1.0.
```
svick: Are you saying the name should be `ChargeLevelPercentage`, but return a value between 0.0 and 1.0? I think that would be wrong. "Percentage" specifically refers to a fraction multiplied by 100, i.e. the value would have to be between 0.0 and 100.0.

PR #72, opened 2018-10-25
Add 2018/IAsyncEnumerable

terrajobst:


2018/IAsyncEnumerable/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+* We believe we need an `IAsyncDisposable` in order to model releasing resources
+  in an asynchronous fashion, which is often needed, for example to close file
+  handles. The compiler will use this in order to model `try/finally` in
+  iterators.
+
+## Notes
+
+* We're using `ValueTask` and `ValueTask<T>` to improve performance in cases
+  where the operation completes asynchronously.
```
svick: Should this be "synchronously"?

```suggestion
  where the operation completes synchronously.
```

stephentoub: No, asynchronously is correct.  We could handle being allocation-free for synchronous completions just by using cached `Task<bool>`s for true and false, for example.  But `ValueTask<bool>` let's use be amortized-allocation free for asynchronously completion as well.

2018/IAsyncEnumerable/README.md
```diff
@@ -0,0 +1,34 @@
+# IAsyncEnumerable
+
+Status: **Needs more discussions** |
```
stephentoub: > Needs more discussion

My take away is that we're essentially locked on the plans unless someone comes up with a better idea for something, in which case we can discuss and pivot as needed.  Is that not right?

2018/IAsyncEnumerable/README.md
```diff
@@ -0,0 +1,34 @@
...
+  can only be used in an async context.
+
+## Next actions
+
+* We need to talk more about cancellation
+    - Right now, it's ambient state that the implementer has to store on the
+      the instance that implements `IAsyncEnumerable<T>`, which doesn't seem
+      right as the state belongs to the enumerator, not the enumerable. Which
+      means single use enumerable is fine, multi-use isn't.
+    - The problem is there no easy way to pass it
```
stephentoub: I think we're basically going with the current model unless someone comes up with something better.

2018/IAsyncEnumerable/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+* We're using `ValueTask` and `ValueTask<T>` to improve performance in cases
+  where the operation completes asynchronously.
+* `ConfigureAwait` is supported with some machinery in `S.R.CS`. The support is
+  the same
+    - We'll add an extension method to S.T.Task.TaskExtension
+* It should be OK for a type to only implement `IAsyncDisposable` if the type
+  can only be used in an async context.
+
+## Next actions
```
stephentoub: A few more outcomes:
- Rather than Stream implementing DisposeAsync as virtual, we'll add a protected virtual FlushFinalAsync, and Dispose will call FlushFinalAsync and then Dispose.   FlushFinalAsync will just call FlushAsync in the base implementation, but a type like DeflateStream can override it to do whatever trailer work it needs to.  After a call to FlushFinalAsync, Stream.Dispose should only have synchronous work to do.  We'll use the same pattern on TextWriter.
- We'll rename ManualResetValueTaskSourceLogic to ManualResetValueTaskSourceCore.

2018/IAsyncEnumerable/README.md
terrajobst: Singular?

terrajobst: > My take away is that we're essentially locked on the plans unless someone comes up with a better idea for something, in which case we can discuss and pivot as needed. Is that not right?

We need to close on the cancellation token. My impression from the meeting was that this aspect wasn't resolved.

stephentoub: > We need to close on the cancellation token. My impression from the meeting was that this aspect wasn't resolved.

We discussed several options:
1. Having IAsyncEnumerator.MoveNextAsync take a cancellation token.  We quickly agreed this doesn't work well (changing the cancellation token mid-enumeration leads to lots of issues in implementations, the enumeration is logically one operation to be canceled or not, how does the token get into the body of an iterator anyway, how do you pass in the token, etc.)
2. Having IAsyncEnumerable.GetAsyncEnumerator take a cancellation token.  We agreed in theory there are nice aspects of this, but that it also doesn't work very well (how does the token get into the body of an iterator without significant new language features that are specialized to CancellationToken, how do you nicely provide the cancellation token to the await foreach, etc.)
3. Having a type expose a GetAsyncEnumerator (not the interface implementation) that accepts a CancellationToken and returns an IAsyncEnumerator.  There's nothing stopping this from happening today, but it doesn't influence the proposal because it doesn't impact the interface, and it's not directly await foreach'able, which means a consumer either manually uses it or the language needs to add support for await foreach'ing an async enumerator, which has issues around lifetime management (e.g. the await foreach didn't construct the enumerator, should it dispose it?)
4. The current plan of record, which is that you simply store the CancellationToken in the enumerable.  If you're writing an async iterator, that just means taking the CancellationToken as an argument, e.g.
```C#
await foreach (var item in GetDataAsync(token)) { … }
```
If you want to get multiple enumerators from the single enumerator you get back from calling that method, each of those will share the same token.  If you want each to have its own token, you have a separate method that supports canceling just that one consumer, e.g. an AsAsyncEnumerable(CancellationToken) method on whatever it is you're pulling data from, or a WithCancellation(CancellationToken) method that produces a new enumerable with the token embedded, etc.  That method would either only cancel the consumption (which is what you'd want if you were, for example, reading from a channel), or it would need to have knowledge of and internals access to whatever the original enumerable so it could interact with it (as would be the case with something like EF).

So, from my perspective, while there may be discussion remaining about what patterns we want to espouse, whether there are additional language features that would be beneficial, etc., I don't think it impacts any of the interfaces or core types reviewed.

@bartonjs, @KrzysztofCwalina, @MadsTorgersen, @agocke, am I misremembering?  Maybe I'm just hoping it was resolved :smile:

omariom: Can it be that the issues with the design caused by close following the shape of sync enumerable?

I think you need to gather as many cases where async enumerables/iterators will be created/ implemented - like Channels, DB access, files, streams returned from SignalRs etc - and derive what is needed for them. I can be one interface or many (even more than two, if required), one method or many.

I may be telling obvious things and you have already done it. peace :v:

stephentoub: > Can it be that the issues with the design caused by close following the shape of sync enumerable?

Not from my perspective. We've evaluated many shapes over many years against many needs. The issues discussed at the review were rehashes of discussions we've had multiple times in the past (sometimes with different people, sometimes with the same), other than the issues i raised around Stream which have little to do with interface shape and more to do with the difficulties inherent to augmenting existing types (e.g. you can't add abstract methods, you need to be cognizant of existing derived implementations, etc.)

divega: @stephentoub Re this argument:

> 3. Having a type expose a GetAsyncEnumerator (not the interface implementation) that accepts a CancellationToken and returns an IAsyncEnumerator. There's nothing stopping this from happening today, but it doesn't influence the proposal because it doesn't impact the interface, and it's not directly await foreach'able, which means a consumer either manually uses it or the language needs to add support for await foreach'ing an async enumerator, which has issues around lifetime management (e.g. the await foreach didn't construct the enumerator, should it dispose it?)

I know I am the one that brought it up in the meeting, but afterwards I came to think about it as a weak argument. If foreach (and await foreach) started supporting enumerators directly, then I think it would be fine for it to always dispose the enumerator. During the meeting I failed to make this connection, but this is actually pretty consistent with a using statement that receives an IDisposable instance created in user code, either between the parenthesis or elsewhere.

Not saying that this invalidates the arguments against GetAsyncEnumerator taking a cancellation token in the interface (at least as long as we don't have a good solution for flowing the cancellation token all the way from the foreach to the body of the iterator), or that this is enough justification to make foreach and await foreach support working with enumerators. GetAsyncEnumerator isn't a nicely named/friendly API I would like our customers to have to start using on a daily basis.

PR #76, opened 2018-11-09
Add notes for UtfString-related APIs

terrajobst:
/cc @GrabYourPitchforks @migueldeicaza

### DIFF
--- a/2018/System.Utf8String/Session1.md
+++ b/2018/System.Utf8String/Session1.md
@@ -1,4 +1,4 @@
-# UTF8 String
+# UTF8 String (Session 1)
 
 Status: **Needs more work** | 
 [API](https://github.com/dotnet/corefx/issues/30503) |
--- a/2018/System.Utf8String/Session2.md
+++ b/2018/System.Utf8String/Session2.md
@@ -1,0 +1,22 @@
+# UTF8 String (Session 2)
+
+Status: **Needs more work** | 
+[API](https://github.com/dotnet/corefx/issues/30503) |
+
+## UnicodeScalar
+
+* `String.GetScalarAt(): UnicodeScalar`
+    - Should be a static on `UnicodeScalar`, as it's tricky
+* `String.GetScalars()` should be an instance `String`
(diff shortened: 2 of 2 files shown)


vectorix:
Rune is not a precise name unless it actually describes a letter of the runic alphabet. Otherwise it becomes an overloaded term.

Ironically, a rune is already defined to be one of the following Unicode letters in the standard: https://www.unicode.org/charts/PDF/U16A0.pdf

Why not rename UnicodeScalar to Hangul? It would be just as imprecise and fun to use in Korea.

benaadams:
Rune was the term used by the creator of Utf-8 when creating Utf-8 https://www.cl.cam.ac.uk/~mgk25/ucs/utf-8-history.txt

He's now works on golang; which is why golang uses rune https://blog.golang.org/strings

terrajobst:
Ben is correct, that’s the rationale.

psibernetic:
Going to agree with the logic for the name, the creator named it that IN an implementation of a mainstream language. Other than getting his direct opinion on the matter, this makes sense to me.

tannergooding:
> Rune was the term used by the creator of Utf-8 when creating Utf-8 https://www.cl.cam.ac.uk/~mgk25/ucs/utf-8-history.txt
>
> He's now works on golang; which is why golang uses rune https://blog.golang.org/strings
>
> The Unicode block is called Runic not Rune, and they are Runic letter(s), not Rune letter(s)

@benaadams, I don't think any of that really matters. Lots of people use "bad" names for prototypes, etc.

I think that what really matters is that the current/latest Unicode Term Glossary does not use or define "Rune". It defines and talks in terms of [Code Point](http://www.unicode.org/glossary/#code_point) and [Code Unit](http://www.unicode.org/glossary/#code_unit).

Additionally, if you search through the current Unicode Specification (http://www.unicode.org/versions/Unicode11.0.0/UnicodeStandard-11.0.pdf) for the term "rune", there are a total of 37 hits (starting on page 337) and every hit is talking about the Runic Alphabet and the individual characters ("runes") that comprise it.

PR #78, opened 2018-11-09
Add notes for round 3 of GPIO

terrajobst:


2018/System.Device.Gpio3/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Status: **Needs more work** | 
+[API Ref](System.Device.Gpio3.md) |
+[Video](https://www.youtube.com/watch?v=UZc3sbJ0-PI)
+
+## Notes
+
+* `AddCallbackForPinValueChangedEvent`/`RemoveCallbackForPinValueChangedEvent` is weird.
+    - Can we simplify the arguments?
+    - Can we provide a lightweight struct that allows regular event subscription
```
joperezr: I think the proposal you show looks great. During the review I believe we agreed that we would start with the Register/Unregister approach and if we get feedback we can always come back to do this instead.

2018/System.Device.Gpio3/README.md
```diff
@@ -0,0 +1,57 @@
...
+* `WaitForEvent` is lossy
+* `GpioController` shouldn't have a finalizer
+
+## I2c
+
+* `uint` -> `int`
+* The interface should be abstract class
+* Consider hiding the Unix type and replace it with a factory on the the abstract class
+* Ideally though, you make it consistent with `Gpio`
+* Instead of implementing reading/writing yourself, consider exposing a `Stream`. You might
```
joperezr: I chatted with @stephentoub and Stream isn't a very good candidate in our case. Instead we will have the following overloads for reading and writing:
```c#
// Reading
byte ReadByte(int address);
void Read(int address, Span<byte> buffer);

// Writing
void WriteByte(int address, byte data);
void Write(int address, Span<byte> data);
```

stephentoub: > Stream isn't a very good candidate

The specific thing that makes it undesirable is the `int address`, with the use cases @joperezr described to me being random-access with arbitrary register addresses, rather than being a forward-reading stream where each read moves to the "next" position.

I also asked about whether `Read(int address, Span<byte> buffer)` should return the number of bytes read, with a max being `buffer.Length`.  But unlike in Stream where that's the behavior because it's expected you may only be able to read fewer bytes than were requested, in these scenarios, that's very unexpected; it's expected that the caller knows exactly how many bytes are needed and available and calls with a span of exactly that length... if fewer bytes are available, it's exceptional.

terrajobst: Sounds good to me. Given the note says 'consider' we can merge this as-is, as you considered it, but decided against it.

PR #85, opened 2019-01-15
Adding notes for Index and Range

terrajobst:


tarekgh:
https://github.com/dotnet/corefx/issues/34076

2019/System.Index/README.md
```diff
@@ -0,0 +1,55 @@
...
+* `Index` has an implicit conversion to `int`. However, it will fail for
+  negative numbers. We normally don't allow implicit conversions to fail;
+  however, we believe the developer experience is virtually identical as all
+  methods accepting an `int`-based index will throw an
+  `ArgumentOutOfRangeException` when being called with a negative number. Now
+  the call site will fail with the conversion, but it's the same line of code in
+  the developer's program.
+    - We should `ArgumentOutOfRangeException` from the implicit operator (as
+      opposed to `InvalidCastException`)
+* Do the types `Index` and `Range` any value over conventions for method/indexer parameter list?
```
ericstj: missing verb here.

2019/System.Index/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+Code like this could work:
+
+```C#
+x.Slice(start..end);
+```
+
+There would be limitations though:
+
+* The `^` syntax wouldn't be supported
```
benaadams: Why? It could be supported on types with a .Length property

If 
```csharp
x.Slice(start..end);
```
Can become 
```csharp
x.Slice(start, end - start);
```
Then
```csharp
x.Slice(start..^end);
```
Can become
```csharp
x.Slice(start, x.Length - end - start);
```

terrajobst: > Why? It could be supported on types with a .Length property

Excellent point; I haven't thought about that. I'll update the comment.

2019/System.Index/README.md
```diff
@@ -0,0 +1,55 @@
...
+* Use of `Index` and `Range` types wouldn't be supported
```
benaadams: Again, could deconstruct them into the parameters? (using `.Length` in a ternary for from end)

terrajobst: Maybe; I'm curious how bad the code spit would be for handling all cases. If perf is critical, it might still be a problem. However, it's worth calling out that it's at least possible.

benaadams:
```csharp
int start = range.Start.FromEnd ? x.Length - range.Start.Value : range.Start.Value;
int end = range.End.FromEnd ? x.Length - range.End.Value : range.End.Value;
x.Slice(start, end - start);
```
Stolen from [`Span<T> this[Range range]`](
https://github.com/dotnet/coreclr/blob/3f0903598d0e269e6aaf8e237b47269abe4ab7e8/src/System.Private.CoreLib/shared/System/Span.Fast.cs#L171-L179)

benaadams: Also from same place
```
int actualIndex = index.FromEnd ? x.Length - index.Value : index.Value;
x[actualIndex]
```

2019/System.Index/README.md
```diff
@@ -0,0 +1,55 @@
...
+    - Performance issues (see below)
+
+## Convenient use of high-performance APIs
+
+Indexers and slice methods are sometimes performance critical, especially for
+low-level types, for example, `Span<T>`. We generally don't believe we'll be
+able to get them to parity with `int`-based overloads. We can make them additive
+in the sense that we can add overloads, but this also means callers cannot use
+language syntax. An alternative would be to allow attributing the method (or
+parameters) that allows the caller to use syntax at the call site.
```
KrzysztofCwalina: BTW, in addition to the overhead of precondition checks in Range, Span<T>.Slice(int) is a much better method to implement [X..] (left bound range). And so, I would love if the attribute allowed such syntax to bind to single parameter methods and in case of ranges bound from both sized calling the two int overload. In addition, if the feature integrated with the Length property as well, [1..^5] could bind to span.Slice(1, span.Length - 5 - 1)

terrajobst: I've updated the document to call this out.

PR #87, opened 2019-01-22
Add second round of notes for Index and Range

terrajobst:


2019/System.Index/README.md
```diff
@@ -24,6 +24,58 @@ Status: **Review not complete** |
...
+  developer experience doesn't suffer meaningfully.
+
+### Index
+
+We need this method:
+
+```C#
+struct Index
+{
+    static readonly Index Start = new Index(0, false);
```
jkotas: Should these be properties instead of readonly fields? Properties are much easier for codegen to optimize than ReadOnly fields containing structs.

2019/System.Index/README.md
```diff
@@ -24,6 +24,58 @@ Status: **Review not complete** |
...
+    static readonly Index End = new Index(0, true);
+    Index(int value, bool isFromEnd = false) {}
+    static Index FromStart(int value) {}
+    static Index FromEnd(int value) {}
+    bool IsFromEnd { get; }
+    int GetOffset(int length) {}
```
jkotas: What does this do? There is no mention of GetOffset method in https://github.com/dotnet/corefx/issues/34076

terrajobst: It's for the receiver of the index so that they can avoid writing code like this:

```C#
public T this[Index index]
{
    get
    {
        var offset = index.IsFromEnd ? Length - index.Value : index.Value;

        return this[offset];
    }
}
```

And instead just write this:

```C#
public T this[Index index]
{
    get
    {
        var offset = index.GetOffset(Length);
        return this[offset];
    }
}
```

2019/System.Index/README.md
terrajobst: > Properties are much easier for codegen to optimize than ReadOnly fields containing structs.

Sounds good to me, will change.

PR #89, opened 2019-01-31
Add UTF8 Notes

terrajobst:


GrabYourPitchforks:
@terrajobst I think the general consensus was on the method names being `ConvertToUtf16` and `ConvertFromUtf16`?

terrajobst:
@ahsonkhan corrected both of our understanding and it's now correct :-)

PR #91, opened 2019-02-12
JSON serialization

terrajobst:


2019/System.Text.Json-Serialization/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+* We're shooting for an MVP in the .NET Core 3.0 timeframe, mostly due to
+  time constraints but also because baking a serializer takes time
+* We should compare & contrast usability
+* "What's the percentage of customers we want to be happy with the built-in JSON
+  serialization."
+* We shouldn't return `Span<T>`; we should let the caller pass in the buffer to
+  fill. This likely also requires `out` parameters to indicate how many bytes
+  were written.
+* The APIs returning `ValueTask` should probably return `Task`
```
ahsonkhan: Also add, we should re-order the parameters so that the object is first (as source) and the destination is second.
For these two APIs, PipeWriter and Stream should be second.

```C#
public static ValueTask ToJsonAsync(this PipeWriter writer, object value, JsonConverterOptions options = null, CancellationToken cancellationToken = default);

public static ValueTask ToJsonsync(this Stream stream, object value, JsonConverterOptions options = null, CancellationToken cancellationToken = default);
```

2019/System.Text.Json-Serialization/README.md
```diff
@@ -0,0 +1,31 @@
...
+[Video](https://www.youtube.com/watch?v=I_CNwnkNDNA)
+
+## Notes
+
+* We're shooting for an MVP in the .NET Core 3.0 timeframe, mostly due to
+  time constraints but also because baking a serializer takes time
+* We should compare & contrast usability
+* "What's the percentage of customers we want to be happy with the built-in JSON
+  serialization."
+* We shouldn't return `Span<T>`; we should let the caller pass in the buffer to
```
ahsonkhan: + Make them regular static methods (rather than extension methods).

2019/System.Text.Json-Serialization/README.md
jkotas: What is the dependency graph going to look like for these?

This makes `System.Text.Json` to depend on `System.IO.Pipelines`. `System.Text.Json` is inbox, `System.IO.Pipelines` is not inbox today.

ahsonkhan: > What is the dependency graph going to look like for these?

That is a great question. We need to figure out the layering. If System.Text.Json depends on System.IO.Pipelines (which it would have to if we were to provide PipeReader/Writer overloads), it would force S.IO.Pipelines to be inbox. Is that even feasible?

cc @davidfowl, @pakrym, @steveharter

davidfowl: You mean be part of Microsoft.NETCore.App? I think it should be in box and I was actually going to suggest it be in box outside of this conversation.

steveharter: Also on the System.Text.Json packaging we will likely need 3 build flavors:
1) NETCore.App inbox
2) .NET Framework package
3) netstandard package

Today we do 1 and 3.

The reason for 2 is the performance of using System.Reflection instead of IL.Emit causes the serializer to be much slower.

ahsonkhan: The netstandard package targets netstandard2.0 which is supported by netfx 4.7.2+
Why do we need a netfx specific package? Shouldn't the nestandard one be sufficient?

jkotas: > Today we do 1 and 3.

I do not think so. The "netstandard" package is a special source code only package. It is not a regular nuget package with binaries.

steveharter: What will .NET Framework use -- the source code only package? IL Emit is not supported by netstandard2.0.

bartonjs: > What will .NET Framework use -- the source code only package?

Yep.  Our current plan is that only netcoreapp gets compiled assets for System.Text.Json

davidfowl: OK so lets settle on this. System.IO.Pipelines will be in box in 3.0. We can file an issue to raise awareness in case there's pushback.

terrajobst: Even if we put pipelines in box, I don't think the JSON component should depend on pipelines. Pipelines could depend on JSON though.

davidfowl: No that’s wrong. I think that’s not right. Let’s discuss this outside of GitHub through. Pipelines should not become a dumping ground for all integrations with it.

PR #95, opened 2019-03-15
More design notes on pipelining

terrajobst:


2019/System.IO.Pipelines/LowLevelPipelines.md
```diff
@@ -44,4 +44,27 @@ forcing consumers to call different APIs.
...
+{
+    public virtual bool CanXxx { get; }
+
+    // Read
+    public virtual ValueTask<ReadResult<byte>> ReadAsync(CancellationToken cancellationToken = default);
+    public virtual void AdvanceTo(SequencePosition consumed, SequencePosition examined);
+
+    // Write
+    public virtual Memory<byte> GetMemory(int sizeHint = 0);
+    public virtual Span<byte> GetSpan(int sizeHint = 0);
```
halter73: We discussed renaming sizeHint to minSize with 0 still being a special case meaning the same as 1.

2019/System.IO.Pipelines/LowLevelPipelines.md
```diff
@@ -44,4 +44,27 @@ forcing consumers to call different APIs.
...
+
+Rough proposal:
+
+```C#
+partial class Stream
+{
+    public virtual bool CanXxx { get; }
+
+    // Read
+    public virtual ValueTask<ReadResult<byte>> ReadAsync(CancellationToken cancellationToken = default);
```
halter73: We discussed renaming this to something else like ReadBufferedAsync or PeekAsync.

2019/System.IO.Pipelines/LowLevelPipelines.md
```diff
@@ -44,4 +44,27 @@ forcing consumers to call different APIs.
...
+    public virtual void AdvanceTo(SequencePosition consumed, SequencePosition examined);
```
halter73: It would be nice to more clearly associate AdvanceTo with reading and Advance with writing now that they're on the same type.

2019/System.IO.Pipelines/LowLevelPipelines.md
```diff
@@ -44,4 +44,27 @@ forcing consumers to call different APIs.
...
+  `AsyncBufferedReader<T>` and `AsyncBufferedWriter<T>` types.
+
+## Discussion
+
+Rough proposal:
+
+```C#
+partial class Stream
+{
+    public virtual bool CanXxx { get; }
```
halter73: CanProvideBuffers was the name I heard thrown around the most so far.

2019/System.IO.Pipelines/LowLevelPipelines.md
davidfowl: I like ReadBufferedAsync

2019/System.IO.Pipelines/LowLevelPipelines.md
davidfowl: AdvanceReader?

terrajobst: So what should we name them?

* `AdvanceTo` -> `AdvanceReader`?
* `Advance` -> `AdvanceWriter`?

Both feel odd as there is no such thing as a writer.

halter73: I like the suggested names. Maybe there's an argument for putting "Write" in the names of GetMemory/GetSpan so they will be more closely associated with AdvanceWriter. Ultimately, I think it's a bit of a struggle to put both reading and writing APIs on the same type once you have much more than the basic ReadAsync/WriteAsync method.

terrajobst: Done

PR #105, opened 2019-08-14
Add notes for Task Dialog APIs

terrajobst:


2019/System.Windows.Forms/README.md
```diff
@@ -0,0 +1,43 @@
...
+  `TaskDialogPage.Created` and `TaskDialogPage.Destroyed` we should align the
+    - Investigate what naming makes most sense across the entire code, maybe do
+      a small user study naming.
+* Rename `instruction` to `mainInstruction`
+* Rename `title` to `caption`
+* Can we remove `TaskDialogStandardIcon` and replace the enum values with static
+  fields on `TaskDialogIcon`. Also instead of implicit operators, can we just
+  have a constructor that takes an `Icon`?
+    - Same for `TaskDialogStandardButton`.
+* Remove `TaskDialogButtonClickedEventArgs`, i.e. make `TaskDialogButton`
```
kpreisser: I think there is something missing at the end of the line (at least I don't understand the "i.e. make `TaskDialogButton`").

* I think it was noted that as "compensation" for removing the `TaskDialogButtonClickedEventArgs`, we would add a `boolean` property to `TaskDialogButton` that allows to specify whether the dialog will close when the button is clicked.

terrajobst: Good catch! Not sure what happened there; I might have accidentally trimmed that bullet when merging my notes with @OliaG's notes. I believe your conclusion is correct and I've updated the doc accordingly.

PR #109, opened 2019-12-03
Add notes for JSON reference handling

terrajobst:


GrabYourPitchforks:
That seems to have fairly captured the conversation. Engineering team may want to clarify a bit what it meant by "escape the _$_". I'll also work on some docs detailing potential security risks of enabling this feature and how to use it safely / successfully inside your own app.

Jozkee:
This is what I understand from "Escape the $": 

* On serialization, when a *JSON property name*, that is either a dictionary key or a CLR class property, starts with a '$' character, we must write the escaped character "\u0024" instead.

* On deserialization, metadata will be digested by using only the raw bytes, so no encoded characters are allowed in metadata; to read JSON properties that start with a '$' you will need to pass it with the escaped '$' (\u0024).

PR #113, opened 2020-03-10
Add round 1 of notes for ASN reader/writer

terrajobst:


2020/03-10-asnreader-writer/README.md
```diff
@@ -0,0 +1,29 @@
...
+**In Progress** |
+[Spec](https://github.com/dotnet/designs/pull/93) |
+[Video](https://youtu.be/PtiTXxroJT4)
+
+## Round 1
+
+* We're reading data that is encoded in that format, as opposed to the
+  description language itself. In XML speak, we're building an
+  `XmlReader`/`XmlWriter` as opposed to `XmlSchema`.
+* Why only a `ref struct` reader? Because `ref struct` writers are error prone
```
bartonjs:
```suggestion
* Why a `ref struct` reader but no `ref struct` writer? Because `ref struct` writers are error prone
```

The only is ambiguous; the flagship reader is a class :smile:

2020/03-10-asnreader-writer/README.md
```diff
@@ -0,0 +1,29 @@
...
+  to use, see JSON writer
+* `Asn1Tag` looks good as proposed
```
bartonjs: Didn't @GrabYourPitchforks want HasSameClassAndValue to change from instance to static? (and no one else had an opinion?)

2020/03-10-asnreader-writer/README.md
```diff
@@ -0,0 +1,29 @@
...
+  * Should we make it non-disposable? We can make it disposable before we
+      ship.
+  * Should the ctor also take an initial capacity?
+  * `GetEncodedLength()` should throw `InvalidOperationException`, rather than
+      returning `-1`
+  * `ValueEquals()` should be `EncodedValueEquals()`
+* What namespace should this go in?
+  * `System.Protocols.Asn1`?
+  * `System.Protocols` would be the home for protocols that aren't important
+    enough to be promoted elsewhere (such as  XML, JSON, HTTP).
```
SteveSyfuhs: It's an encoding though, so shouldn't it go under System.Encoding?

bartonjs: There's not a `System.Encoding`.  There's `System.Text.Encoding`; but that's a class for string<->byte.

2020/03-10-asnreader-writer/README.md
terrajobst: Did he? I can add that.

GrabYourPitchforks: Eh, it's just a weak preference. IMO when _left_ and _right_ are swappable that seems like a better candidate for a static method than an instance method.

2020/03-10-asnreader-writer/README.md
```diff
@@ -0,0 +1,29 @@
...
+* `AsnEncodingRules` looks good as proposed
+* `AsnWriter`
+  * Should we make it non-disposable? We can make it disposable before we
+      ship.
+  * Should the ctor also take an initial capacity?
+  * `GetEncodedLength()` should throw `InvalidOperationException`, rather than
+      returning `-1`
+  * `ValueEquals()` should be `EncodedValueEquals()`
+* What namespace should this go in?
+  * `System.Protocols.Asn1`?
```
filipnavara: There's already `System.Text`, how about `System.Binary.Asn1`? Personally I am not big fan of "Protocols" since I think of it more as a "data format" which is used by "protocol"(s).

GrabYourPitchforks:
> GrabYourPitchforks approved these changes on behalf of dotnet/fxdc 4 hours ago

Cool, I guess I speak for the team now. 😎

2020/03-10-asnreader-writer/README.md
terrajobst: OK, so let's as is then

PR #122, opened 2020-06-16
Add round 2 for Cbor

terrajobst:


eiriktsarpalis:
Probably also needs mentions of special exception replacing FormatException.

terrajobst:
Good point. Fixed by b0028ea.

PR #123, opened 2020-06-23
Networking abstractions: Add notes for round 3

terrajobst:
What are the next steps?

@scalablecory @davidfowl @halter73


scalablecory:
> What are the next steps?

4th round :/

- We need to finish reviewing the "Additional APIs" section.
- We need to decide on virtualness/throwiness of Connection.Stream and Connection.Pipe. (@davidfowl to provide more feedback)

PR #127, opened 2023-11-08
Add review notes for 11/07/2023

teo-tsirpanis:


bartonjs:
Looks like this finally/just got posted by the bot that normally does it.  Thanks for trying to pick up the slack, though.

terrajobst:
@teo-tsirpanis sorry, I didn't even see the PR.

For context, we don't generally create these notes by hand. They are generated from apireview.net and are automatically aggregated by using the live stream + the GitHub comments that happened during that time.

